PR #17, opened 2017-01-24
Add support for clean urls

firstred:


kpodemski:
I'm wondering what about modules (like my blog) which are based on clean urls too, did you test that? :)

firstred:
Yep, have tested code for that. There was a conflict with the `rewrite` param with some blogs.

Gotta test non-SEO friendly URLs, but with friendly URLs seems to work.

kpodemski:
Cool. Good job 👍

PR #22, opened 2017-01-27
Code styling, avoiding annoying warnings in editor

kpodemski:


firstred:
Awesome!

Could you change the property blocks to something like this:
```php
@var Cart $cart
```

This way the IDE knows what a variable is.

kpodemski:
@firstred done, i guess we don't need that in one-liners like $error Controller errors?

firstred:
Does sublime recognize those variables? It's about making things easier for devs afterwards.

I often choose the variables that can otherwise not be inspected. Those have a higher priority atm.

kpodemski:
@firstred Sublime is not an IDE as you may know ;P

firstred:
It doesn't have a plugin to make those things easier?

(I should really check it out one day, haha)

classes/controller/FrontController.php
```diff
@@ -954,7 +973,7 @@ protected function geolocationManagement($defaultCountry)
                                 $this->context->smarty->assign(
                                     [
                                         'restricted_country_mode' => true,
-                                        'geolocation_country'     => $record->country_name
```
firstred: Our code style is aligned arrays. This would revert that.

classes/controller/FrontController.php
```diff
@@ -30,37 +30,41 @@
...
     /**
      * @deprecated Deprecated shortcuts as of 1.5.0.1 - Use $context->cart instead
-     * @var $cart Cart
+     *
+     * @var Cart $cart
      */
     protected static $cart;
 
-    /** @var array Controller errors */
+    /** @var $errorsarray  */
```
firstred: `@var array @errors`

classes/controller/FrontController.php
kpodemski: is "your code style" somehow documented? or named? it's symfony? psr-x?

firstred: Yep, fully documented: https://github.com/thirtybees/ThirtyBees/blob/master/CONTRIBUTING.md

firstred:
Gracias!

PR #48, opened 2017-01-31
Change font color of .notifs_icon_wrapper to #000

inetbiz:
The previous color was a darker orange on a light orange background wrapper.


firstred:
Good catch! Although it might be a bit harder to fix. You have changed the CSS whereas the SCSS file should be changed, otherwise this change would immediately be overwritten by another. I haven't found the correct selector to change the badges, but it should be in: https://github.com/thirtybees/ThirtyBees/blob/master/admin-dev/themes/default/sass/schemes/admin-theme-thirtybees.sass

inetbiz:
@firstred Ack, I'll see if I can hunt that down then.

inetbiz:
```/admin-dev/themes/default/sass/vendor/bootstrap-sass/_badges.scss``` But I do not know where it gets its variable on line **13**?

firstred:
I think from a `_variables.scss` file

inetbiz:
@firstred please review?

firstred:
Wrong file, the default variable is being overwritten.

PR #114, opened 2017-02-21
Add missing properties to Product class.

Hailong:
It would cause errors without these two properties in place when loading the product edit page from backend office.

Note that the error doesn't happen with sample data, while I can reproduce it after I copy over the products and categories tables from an existing PS1.6 DB to TB1.0.0 DB

### DIFF
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -214,6 +214,10 @@ class ProductCore extends ObjectModel
     public $pack_stock_type = 3;
     // @codingStandardsIgnoreEnd
 
+    public $packItems;
+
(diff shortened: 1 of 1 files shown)


firstred:
Thanks! There was a bug in the magic setter method in the ObjectModel class on the master branch, but nonetheless we need these two properties defined first.

Hailong:
Thanks @firstred , I'm happy to help in TB. And our project has been started based on TB.

firstred:
Awesome! That's great to hear. 

We could use all the help we can get right now, so thank you for your contributions.

PR #115, opened 2017-02-22
Bugfixes for full page cache

jankawulok:
Fix incorrect CacheApcu class name in AdminPerformanceController and correct passing $hookName keys of associative array in Controller


firstred:
Awesome! Thank you so much!

PR #140, opened 2017-03-10
Fix customer's address can't be added/edited by admin.

TrueXakeP:
There was a PHP warnings and not all fields was shown in edit address form.


firstred:
Thanks a lot for your contribution! 
We will have a look at it in Monday, when we can take the time to thoroughly check everything.

firstred:
Awesome. Seems to really solve the issue. Thanks a lot!

PR #144, opened 2017-03-13
Url rewrite improvements

firstred:


kpodemski:
What is the reason of moving urls to DB? all i can see now is problems related to this change

firstred:
> What is the reason of moving urls to DB? all i can see now is problems related to this change

Yeah, it's a pretty new feature. We're going to optimize it even further in the coming days.

It works for WordPress, so I don't see any reason why it wouldn't work for thirty bees.

firstred:
Can't update db schema in PATCH versions

PR #162, opened 2017-03-29
File based ShopUrl

Traumflug:
Here we go. This series of commits does:

- Enable *automatic* shop domains. Using them never redirects to the "right" domain, it just loads the shop as-is (1 commit).
- Replace database table ```shop_url``` with file based storage. File is config/shop.inc.php. This is not only faster due to fewer database acesses, it also makes shop domains and URLs easily accessible to shell scripts, in particular shop synchronisation scripts (25 commits, each of them can be tested by running a shop).
- Fix a bug which allowed to delete the main URL of a shop. Deleting main URLs is a bad idea for obvious reasons.

Rebased to current master, tested many many times and found to work flawlessly.

Forum discussion: https://forum.thirtybees.com/topic/76/safe-maintenance-strategy-shop-cloning/14

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ admin-dev/themes/*/cache/*
 
 config/settings.inc.php
 config/settings.old.php
+config/shop.inc.php
 config/xml/*
 !config/xml/themes/default.xml
 
--- a/classes/Configuration.php
+++ b/classes/Configuration.php
@@ -431,6 +431,14 @@ class ConfigurationCore extends ObjectModel
                 static::$_cache[static::$definition['table']][$lang]['global'][$row['name']] = $row['value'];
             }
         }
+
+        // Adjust automatic values.
+        if (static::get('PS_SHOP_DOMAIN') === '*automatic*') {
+            static::set('PS_SHOP_DOMAIN', $_SERVER['HTTP_HOST']);
+        }
+        if (static::get('PS_SHOP_DOMAIN_SSL') === '*automatic*') {
+            static::set('PS_SHOP_DOMAIN_SSL', $_SERVER['HTTP_HOST']);
+        }
     }
 
     /**
@@ -669,6 +677,14 @@ class ConfigurationCore extends ObjectModel
                     );
                 }
             }
+        }
+
+        // Adjust automatic values.
+        if ($key === 'PS_SHOP_DOMAIN' && in_array('*automatic*', $values)) {
+            $values = $_SERVER['HTTP_HOST'];
+        }
+        if ($key === 'PS_SHOP_DOMAIN_SSL' && in_array('*automatic*', $values)) {
+            $values = $_SERVER['HTTP_HOST'];
         }
 
         Configuration::set($key, $values, $idShopGroup, $idShop);
--- a/classes/ObjectFileModel.php
+++ b/classes/ObjectFileModel.php
@@ -117,6 +117,30 @@ abstract class ObjectFileModelCore extends ObjectModel
     }
 
     /**
+     * Gets one or all record(s).
+     *
+     * @param bool $id Record to get. All records if NULL.
+     *
+     * @return array Array with record or array with all record arrays.
+     *
+     * @since   1.1.0
+     * @version 1.1.0 Initial version
+     */
+    public static function get($id = NULL)
+    {
+        global $shopUrlConfig;
+
+        if (!is_array($shopUrlConfig)) {
+            // foreach() loops don't like NULL.
+            return [];
+        } else if ($id) {
+            return $shopUrlConfig[$id];
+        } else {
+            return $shopUrlConfig;
+        }
+    }
+
+    /**
      * Writes the whole objects array as-is to it's file. Should be executed
      * after any permanent change to that array.
      *
--- a/classes/ObjectFileModel.php
+++ b/classes/ObjectFileModel.php
@@ -219,6 +219,67 @@ abstract class ObjectFileModelCore extends ObjectModel
         return $result;
     }
 
+     /**
+      * Updates the current object in the file based storage.
+      *
+      * @param bool $nullValues Ignored, for compatibility with ObjectModel.
+      *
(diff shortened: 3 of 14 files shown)


firstred:
:open_mouth: You did it!

Nice

firstred:
It goes on the 1.0.1 list if possible. We have decided to break SemVer this time anyway, because it's a special occasion. 1.0.1 is going to be the first deadline-free version and will have a special updater.

Traumflug:
Thanks.

How to update an existing installation can be seen here: https://github.com/thirtybees/ThirtyBees/pull/162/commits/0339210cae904670253f378e693bf3d6a92cf642#diff-4dc1babb8598bd71cc36b7f64342cd55R87

It's pretty straightforward. ```(new ShopUrl)->saveShopUrlConfig();``` got replaced in a later commit with ```$this->write();``` or ```(new ShopUrl)->write();```

firstred:
Hmmz, it seems that there's a syntax error:
`PHP Parse error:  syntax error, unexpected '.', expecting ']' in /home/travis/build/thirtybees/ThirtyBees/classes/shop/ShopUrl.php on line 66`

Traumflug:
It's this code:
```PHP
    public static $definition = [
        [...]
        'path'    => _PS_ROOT_DIR_.'/config/shop.inc.php',
        [...]
    ];
```
Apparently doesn't like this dot operator there. Will look up tomorrow on how to write this in PHP 5.x.

Traumflug:
Travis also complains about the new class, ```ObjectFileModel``` missing. The class is certainly there, but being a new class in a new file, one has to clear the classname cache.

firstred:
I think even 7.0 does not know how to handle that. It is a very nice 7.1 feature. Can't wait to drop support for older versions :stuck_out_tongue:

firstred:
Do you think you can get the tests to pass today? I don't mind if it is unstable at this moment. If we can merge it today then there is a big chance that we can pack it with `1.0.1` already. Otherwise we will have to postpone it till `1.1.0`.

If you're having trouble with the tests, let me know. I can try to fix it for you as well.

Traumflug:
I was about to add PHP 5 to my environment for testing myself, but as you're that much in a hurry I simply did a best guess and force-pushed that :-)

Regarding stability ... please test this stuff yourself, too. It changes some very fundamental parts, for example redirection to the main shop domain. Failures in this part result in pages not loading at all or infinite redirects.

... ten minutes later ... all tests appear to succeed with the build, but fail on this one:
```
The command "vendor/bin/codecept run unit,integration -c . --coverage --coverage-xml" exited with 255.
```
That's not verbose enough to allow another guess, will have to get this codecept thing running here. Currently it fails with ```No code coverage driver available```.

firstred:
> but as you're that much in a hurry I simply did a best guess

Cool. If it still doesn't build I'll just merge and fix it. 
I'm the kind of dev who likes to test things in an early stage. We are currently building packages for PS 1.6 => tb 1.0.1 migration which might stick for a while (and we will definitely find a lot of surprises), so it would be cool if we could already have these kinds of features in it :)

Traumflug:
Ah, got it now. Error message is
```PHP Fatal error:  Class 'ObjectFileModel' not found in /home/travis/build/thirtybees/ThirtyBees/classes/shop/ShopUrl.php on line 38```

ObjectFileModel is a new class in a new file. To get this loaded, one has to clear the classes cache after a run without the branch. Apparently Travis doesn't do this.

kpodemski:
Does anyone else except you is interested in this feature? :D I'm not sure what real advantage of this is

firstred:
Cloudways
Softaculous
Installatron
Local dev environment

firstred:
Ahh of course. It's because of the unit tests.

The class files have to be loaded in this PRECISE order: https://github.com/thirtybees/ThirtyBees/blob/master/tests/_support/unitloadclasses.php
The dependencies can't be loaded in a diferent order. You can fix it if you want, but I could also do it for you.

Traumflug:
Thanks for the hint! Did a best guess on that, too. Let's see ...

firstred:
Yeah, it is AspectMock. It can change classes on the fly if needed, but in return you will need to provide a hardcoded list that loads the files (I think it is worth it).

Traumflug:
@kpodemski, the idea is to get shop installations movable and synchronizable. Currently one can't even rename the installation directory without breaking the shop.

Synchronization of shops means that one can change and test changes to a shop locally and while the public shop is running. With that done one can simply run a synchronization script to upload the local installation to the public store. Result: a minimum of downtime and a maximum of safety against typos, misconfigurations and similar glitches. I tried to explain that here: https://forum.thirtybees.com/topic/76/safe-maintenance-strategy-shop-cloning/14

kpodemski:
@Traumflug but what exactly is your problem now? how `shop_url` prevents you from having dev environment with MySQL replication setup? i'm not sure to be honest

Traumflug:
@kpodemski I have no problem.

ThirtyBees (and PrestaShops) problem right now is that one can't synchonize a shop, so one has to fall back to configure, change, update a shop while it's live. A popular strategy (because there is no other choice), still a strategy asking for trouble.

MySQL replication, even if available, doesn't help. Because data like the shop's domain or installation directory is in there, so it has to be changed during synchonization.

And, @kpodemski, if you're interested in discussion features, how about joining the forum? The topic is up there for several weeks now. There was plenty of time to discuss it.

kpodemski:
@Traumflug 

ok then, so basically you would like to have possibility to update `dev.store.com` and have possibility to upload effect to `store.com` without changing `shop_url`?

Traumflug:
"All checks have passed" :-)

firstred:
Merged! Thank you

firstred:
If only we could get the forked upgrader module working we could start testing, but it seems we need a few days extra to get the design flaws (a.k.a. spaghetti code) out of the update module (PS version is called `autoupgrade`)

PR #164, opened 2017-04-01
Remove PS_SHOP_DOMAIN and PS_SHOP_DOMAIN_SSL

Traumflug:
Next step towards shop snychonizability is to get rid of configuration parameters PS_SHOP_DOMAIN and PS_SHOP_DOMAIN_SSL. My impression is, they're deprecated in favor of asking ShopUrl already, this move just wasn't completed, yet.

Anyways, here are 7 commits for removing them entirely, plus one commit removing some disconnected comments I found while searching around.

While there's code establishing backwards compatibility, modules authorizeaim and cronjobs should be fixed, too. Fixes are simple, see commit comment here: https://github.com/thirtybees/ThirtyBees/commit/0614b131da05623c60ba1fe2f74dd23fa53cd626

### DIFF
--- a/classes/shop/ShopUrl.php
+++ b/classes/shop/ShopUrl.php
@@ -189,6 +189,34 @@ class ShopUrlCore extends ObjectFileModel
     }
 
     /**
+     * Get the default shop URL, which is the first defined main URL.
+     *
+     * @return bool|ShopUrl ShopUrl object or false on failure to find one.
+     *
+     * @since   1.1.0
+     * @version 1.1.0 Initial version
+     */
+    public static function getDefaultUrl()
+    {
+        global $shopUrlConfig;
+
+        $result = false;
+        if (is_array($shopUrlConfig)) {
+            foreach ($shopUrlConfig as $id => $url) {
+                // We could also check $url['active'] here, but the GUI
+                // (AdminShopUrlController) disallows to disable the main URL,
+                // so the main URL is always active, unless tweaked manually.
+                if ($url['main']) {
+                    $result = new ShopUrl($id);
+                }
+                break;
+            }
+        }
+
+        return $result;
+    }
+
+    /**
      * @return bool
      *
      * @since   1.0.0
--- a/classes/AdminTab.php
+++ b/classes/AdminTab.php
@@ -2609,10 +2609,10 @@ abstract class AdminTabCore
      */
     protected function warnDomainName()
     {
-        if ($_SERVER['HTTP_HOST'] != Configuration::get('PS_SHOP_DOMAIN') && $_SERVER['HTTP_HOST'] != Configuration::get('PS_SHOP_DOMAIN_SSL')) {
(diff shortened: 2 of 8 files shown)


classes/shop/ShopUrl.php
```diff
@@ -189,6 +189,34 @@ public static function getShopUrls($idShop = false)
...
+     * @since   1.1.0
+     * @version 1.1.0 Initial version
+     */
+    public static function getDefaultUrl()
+    {
+        global $shopUrlConfig;
+
+        $result = false;
+        if (is_array($shopUrlConfig)) {
+            foreach ($shopUrlConfig as $id => $url) {
```
kpodemski: hint: `array_filter`

classes/shop/ShopUrl.php
```diff
@@ -189,6 +189,34 @@ public static function getShopUrls($idShop = false)
...
+     * Get the default shop URL, which is the first defined main URL.
+     *
+     * @return bool|ShopUrl ShopUrl object or false on failure to find one.
+     *
+     * @since   1.1.0
+     * @version 1.1.0 Initial version
+     */
+    public static function getDefaultUrl()
+    {
+        global $shopUrlConfig;
```
kpodemski: why not keep this in Context? using $globals these days?

classes/shop/ShopUrl.php
Traumflug: You commit showing me on how to do this with fewer code is welcome!

classes/shop/ShopUrl.php
Traumflug: Because initially there is no context. ShopUrl::defaultUrl has to work very early.

firstred: Not sure if a global is the way to go then. Let's see if we can find an alternative first.

Traumflug: Well, this global variable was introduced with the previous pull request already. One can change this, of course, but if so, not only in this place. Reasons why I choosed to use a global include:

- I was keen on good caching performance. Content of this global is set by an include_once(), so it's well cacheable code, maximizing performance.
- When starting with class ObjectFileModel I was pretty unsure how well I could encapsulate that stuff, so I started with that global. And left it this way, because it works well enough (for now).
- I'm not aware of a better solution. Which doesn't mean much, because I'm not exactly experienced with PHP and coding for good PHP performance. My most used language is plain C for embedded devices, where we have a compiler.

If somebody can show me on how to to place this variable better, I'm all ears. Code is in ShopUrl.php and ObjectFileModel.php. Showing, like showing actual code, please :-)

firstred: Sure thing. I'm on it!

firstred: A nice thing about C is that you always have the option to free or overwrite the memory. With a PHP global you're fked.

PR #166, opened 2017-04-03
Make the reversal of file based ShopUrl storage complete.

Traumflug:
It looks like the reversal of file based ShopUrl storage wasn't as complete as it should have been, so I made this commit to make the reversal complete.


firstred:
Thank you!

PR #167, opened 2017-04-04
File based ShopUrl, take two

Traumflug:
As at least two people were uncomfortable with using a global variable, so I searched for better solutions. Searching, like implementing five of them. It's only the first step of the commit series and can only change already existing URLs in Advanced Preferences -> Multishop. Even then, content written to the file is just a copy of database content. The other 25 commits will follow after a decision about these five versions of the first commit was found.

Commit 1 is what was there before.  ShopUrls stored in global $shopUrlConfig. Name of that variable hardcoded.

Commit 2 abstracts that a bit and provides a class property for the variable name. Something compiled languages don't allow to do. This should allow to treat multiple instances of distinct storage once this code goes into a common base class similar to ObjectModel.

Commit 3 introduces a subclass for storage. Which means the global variable goes away, just to be replaced by some other global entity, a class.

Commit 4 does almost the same, but with a trait. This makes variable access a bit shorter.

Commit 5 uses include()s inside methods, making $shopUrlConfig a local variable in each method.

Commits 3, 4 and 5 don't work flawlessly, but 3 and 5 might allow to go with workarounds. Caveats are described in the commit message.

Please do your choice, folks. Further refinements are also welcome, of course.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,7 @@ composer.lock
 
 /config/settings.inc.php
 /config/settings.old.php
+/config/shop.inc.php
 /config/xml/*
 /!config/xml/themes/default.xml
 
--- a/classes/Configuration.php
+++ b/classes/Configuration.php
@@ -435,6 +435,14 @@ class ConfigurationCore extends ObjectModel
                 static::$_cache[static::$definition['table']][$lang]['global'][$row['name']] = $row['value'];
             }
         }
+
+        // Adjust automatic values.
+        if (static::get('PS_SHOP_DOMAIN') === '*automatic*') {
+            static::set('PS_SHOP_DOMAIN', $_SERVER['HTTP_HOST']);
+        }
+        if (static::get('PS_SHOP_DOMAIN_SSL') === '*automatic*') {
+            static::set('PS_SHOP_DOMAIN_SSL', $_SERVER['HTTP_HOST']);
+        }
     }
 
     /**
@@ -673,6 +681,14 @@ class ConfigurationCore extends ObjectModel
                     );
                 }
             }
+        }
+
+        // Adjust automatic values.
+        if ($key === 'PS_SHOP_DOMAIN' && in_array('*automatic*', $values)) {
+            $values = $_SERVER['HTTP_HOST'];
+        }
+        if ($key === 'PS_SHOP_DOMAIN_SSL' && in_array('*automatic*', $values)) {
+            $values = $_SERVER['HTTP_HOST'];
         }
 
         Configuration::set($key, $values, $idShopGroup, $idShop);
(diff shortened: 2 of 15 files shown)


Traumflug:
P.S.: Travis tests don't matter for now, it's a code design question rather than code to be merged.

Traumflug:
P.P.S.: if you wonder why there's no attempt to include() a class property, look here: https://bugs.php.net/bug.php?id=27569&edit=2

firstred:
Were you able to restore everything you wanted as well?

Traumflug:
I have all the previous commits, if you mean that. Now I'd like to know onto which of the variants above I should apply and adjust them.

My favorite would be a class level include(), but as this isn't allowed, next best is the abstracted global variable in commit 2. The database is kind of a global variable, too, and having a global doesn't mean one has to access it from everywhere. One can still use accessors.

firstred:
What makes you think an `include` will not work? I have never seen parse errors caused by that.

Traumflug:
See here: https://bugs.php.net/bug.php?id=27569&edit=2

One can put includes in before opening the class definition, but then it's a global variable. One can put that include into a class method, then it's a method-local variable (see commit 5). On the class level, includes are simply not allowed.

firstred:
Oh, now I'm beginning to understand. Outside the class or in the constructor is not an option?

Traumflug:
If I understand the language correctly, a definition outside the class gives a global variable.

Constructor is IMHO no option, because most accesses ask for the bare data, only, so demanding construction of instances each time would be a performance setback. Ignoring the performance setback it's possible, of course.

Traumflug:
@kpodemski, this is the right time to chime in for finding alternatives to that global variable you didn't like last time :-)

firstred:
Singleton?

firstred:
Maybe we can make it behave like currencies. The best way to retrieve a currency modules is by calling `Currency::getInstanceById()`, which allows you to recycle the instance. Could that be an idea?

firstred:
It sounds a bit hacky to me, but this might be the best way to reuse `ObjectModel`s

Traumflug:
I'll give Singletons a try to see how it works out.

Traumflug:
In the 6th or 7th attempt I managed to get rid of the global variable and also without subclasses, traits and such stuff. The trick is to use an accessor and only that. There isn't even class-level storage, so also no need to fill that by instance construction or other inefficient strategies: 

This is the game changing commit: https://github.com/thirtybees/thirtybees/pull/167/commits/f845e61910a76545f71e40f1cd03ae50fc118165
This is the resulting class: https://github.com/Traumflug/ThirtyBees/blob/f845e61910a76545f71e40f1cd03ae50fc118165/classes/ObjectFileModel.php#L124

Traumflug:
Rebase to current master and one bug prohibiting installation fixed.

PR #170, opened 2017-04-09
Improved package build script.

Traumflug:
Having a package build script which does self-destruction of the Git repository is a really bad idea. Imagine what happens if a developer simply tries ...

This one works without destroying the original repository and can be invoked right from the development repo. It also makes the resulting package about 3 MB smaller due to better cleaning of vendor dependencies.

Using it is trivial: run ```./build.sh``` in the development repository, wait a minute and enjoy the readily built package.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,18 +1,123 @@
 #!/usr/bin/env bash
-composer install --no-dev
-find -type d -name tests -exec rm -rf {} \;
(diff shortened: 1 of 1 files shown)


firstred:
Awesome. Lots of improvements. Thanks!

Traumflug:
Glad to see you like it :-)

PR #172, opened 2017-04-11
Misc

Traumflug:
Three commits of lower importance, still too valuable to throw them away.

- _Remove constant Module::CACHE_FILE_MODULES_LIST_: last usage of this constant was removed yesterday, and it was dead code.
- _Remove unrelated comments_: apparently forgotten earlier, ```INSTALL_VERSION``` isn't used at all.
- _Be ignorant about submodules_: this commit actually changes something. To me it looks like thirty bees' development model doesn't match Git's default development model too well. Thirty bees apparently develops modules (and the theme) entirely separately and doesn't track changes there with the main repo. Accordingly, the main repo should ignore changes in submodules. Without this, Git laments about submodule changes all the time and a developer has to take care to not pick up such changes in unrelated commits by accident.

### DIFF
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -1624,17 +1624,6 @@ class ToolsCore
     }
 
     /**
-     * Truncate strings
-     *
-     * @param string $str
-     * @param int    $max_length Max length
-     * @param string $suffix     Suffix optional
-     *
-     * @return string $str truncated
-     */
-    /* CAUTION : Use it only on module hookEvents.
-    ** For other purposes use the smarty function instead */
-    /**
      * getHttpHost return the <b>current</b> host used, with the protocol (http or https) if $http is true
      * This function should not be used to choose http or https domain name.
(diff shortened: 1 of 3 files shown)


firstred:
Hi, thanks for your pull request.

The first two look good, but I don't think we can merge the last one. Especially not in this phase of the project. Our currently workflow is this: we ignore submodules, but use `dirty`, to make sure we can still see the changes. Submodules are listed in the `.gitignore` file, so changes to those can only be pushed to their own repos. If we think the next version of the submodule is suitable to be used for the main software, we then update the commit number the submodule refers to and the latest version of the submodule becomes part of the core software.

Traumflug:
> Our currently workflow is this: we ignore submodules, but use dirty, to make sure we can still see the changes.

Makes sense :-) So I'll keep this commit privately and updated the branch / pull request.

firstred:
Thank you!

PR #185, opened 2017-04-19
Carrier wizard audit

Traumflug:
A series once started to fix this unexpected behavior: https://forum.thirtybees.com/topic/184/carrier-fee-not-editable

... and ending up in an almost complete review. 260 lines of code gone. For example, there was a function for sorting ranges I found only by looking into the code: in practice it's always tagged ```display: none``` For my taste it's a bit pointless to maintain code which never executes. Also removed quite some custom handling and hardened validation. Details in the commit comments. Every stage should give a fully working shop.

About API changes: data sent back remains the same, but some translation texts changed, including removed and added strings.


firstred:
Thanks a lot for your contribution. I think it is too late to merge this into `1.0.1`, but we are going to make sure it can be merged with `1.0.2`.

firstred:
Nope. We can still merge it. (Sub)module updates are fine, too. Thank you!

PR #188, opened 2017-04-21
More file checking stuff

Traumflug:
As promised in the forum yesterday, here are the integrity test fixes, along with a faster, PHP based generation script and an enhanced user interface (missing files.json gets reported instead of assuming everything is fine without that file).

Third commit allows developers not committing directly to master to build packages of their current work.

As always, details and reasons for each step in the commit comments.


firstred:
Very useful. Thank you!

PR #189, opened 2017-04-22
Installer error reporting

Traumflug:
Again cases of too much code :-)

Without this, installation errors generate a generic error message, only. Messages are available, so let's display them! Most of the code was already there, it just had to be kind of enabled.

No change for successful installations, except step descriptions now shown in normal text instead of strikethrough.


firstred:
Very nice and a very important change. Thank you!

PR #193, opened 2017-04-24
Fix hardcoded maximum 3MB image size in webservice image upload.

roband7:


Traumflug:
Nice catch.

Nitpicking: tabs instead of spaces in the new line 1139.

firstred:
No worries. I'll take care of that. Thank you!

PR #194, opened 2017-04-24
Add support for gzip compressed input in the webservice.

roband7:


firstred:
Nice change. Thank you!

PR #195, opened 2017-04-25
community theme rtl support

yaniv14:


Traumflug:
Merges are evil :-)

I think the way to go now is to create a branch based on upstream master with this commit, then to pull-request that. Like (educated guess)
```sh
git checkout upstream/master
git checkout -b rtl-support
git cherry-pick 0558f32
git push origin rtl-support
```
then coming back here for making a new PR with branch 'rtl-support'.

yaniv14:
I've just PR it again under new branch, but I see it fails some checks, which I don't really understand.

PR #196, opened 2017-04-25
added isRtl variable to templates

yaniv14:


Traumflug:
Travis error message:

```The command "eval git submodule update --init --recursive " failed 3 times.```

Just tried myself, this command works here. It smells a bit like a Travis problem. Not sure whether it'll try again on it's own.

PR #209, opened 2017-04-29
community-theme-united was unlinked

inetbiz:
This theme was not linked in core.


firstred:
Correct. It will be included in an alternative way, somewhere in the future.

PR #210, opened 2017-04-29
thirty bees - update frontmatter of debug.tpl

inetbiz:
Also compressed css color on line 98 and removed script type line 165 for html5


firstred:
Thank you!

PR #219, opened 2017-05-07
Pickings from PS 1.7

Traumflug:
While trying to get a PS 1.7 theme running with 30bz I tripped over a couple of commits on PS' 1.7 branch which should be helpful here, too. Each of the 16 commits is an independent topic, please cherry-pick in case not all of them should be taken.


Traumflug:
Nah. Forgot to rebase.

PR #240, opened 2017-05-21
Merge 1.0.x changes into 1.1.x

firstred:


Traumflug:
Personally I'd prefer to rebase 1.1.x to 1.0.x instead of merging. Merging creates a non-linear history, which makes bisecting and other rebases harder. Rebasing creates a linear history and treats 1.1.x like a topic branch.

Another way is to cherry-pick each commit of 1.0.x onto 1.1.x. This creates a linear history on each branch as well, partially duplicating it.

To make rebasing as easy as possible, one can step-rebase. Like
```sh
git rebase 1.0.x~28
git rebase 1.0.x~27
git rebase 1.0.x~26
...
```
There's even a tool for doing this automatically: https://github.com/Traumflug/Teacup_Firmware/blob/master/tools/git-step-rebase . Install this somewhere into PATH, then
```sh
git checkout 1.1.x
git step-rebase 1.0.x
```
With this technique, conflicts are always as small as possible. If one appears, resolve it as usual, `git rebase --continue`, then start `step-rebase` again.

firstred:
Thanks a lot Markus. I think so, too, haha :D

PR #245, opened 2017-05-28
Request for comment: support for PS 1.7 themes

Traumflug:
While this work is far from completed (please do not merge it), I think it's time to show current progress and ask for opinions.

One of the implemented debug measures is to always load the boilerplate template, no matter which template is requested. This will go away later, of course, but simplifies things now.

What works fine:

- Finding and loading templates. PS 1.7 uses `config/theme.yml` to define templates and hooks used and that's what the code currently always uses. This removes the need for a backoffice page for defining these relationships. Such a page can be added later, of course. For now, if a user wants to redefine these relationships she can edit this (human readable) file.

- Detecting PS 1.7 themes. The mentioned config file is unique to PS 1.7 themes, so whenever such a file is seen, FrontController switches to PS 1.7 mode.

- Loading assets (CSS, JS). 30bz code is quite messy here (requests to add a CSS file as well as a JS file gets joined to 'addMedia', then CSS and JS are separated again, just to add them to a list). All this fails in a PS 1.7 theme and this is actually wanted behavior. Wanted, because PS 1.7 themes build all this into a single file at compile time, so just one main CSS and two JS files have to be loaded, always the same path, no matter which controller requests the page.

What works mediocre:

- Dealing with modules. Modules are still PS 1.6 compatibles, of course, so they sometimes rely on a PS 1.6 theme structure.

- Template variables. PS 1.7 changed the set of variables available in templates by some 95%. Almost all objects are gone and the few left are apparently deprecated. Instead, data is forwarded by plain arrays, only. Accordingly, there can't be method calls from within a template. I like this design, a clear distinction between controller logic and view.
Nevertheless, Smarty assignments are spread like confetti all over the code, so it's hard to find what's needed for PS 1.7 templates and what not. So far I did this by calling sibling pages in 30bz and PS 1.7 and comparing the dataset available for the template.

What doesn't work so far:

- Translations in the controller. PS 1.7 translates quite a number of strings in FrontController (and subclasses) and I found no equivalent in 30bz so far.

Ideas:

- Move theme handling out of FrontController into a Theme class. There is a class `Theme` already, but this is a manager for dealing with the set of themes, not for serving the theme in use. With a separated class Theme there could be one for PS 1.6 and one for PS 1.7 themes, FrontController just needs to redirect to the appropriate one.

- One of the significant problems is that code like `$this->context->smarty->assign()` can be found everywhere. Instead of trying to adjust all these assignments, one could try to let them happen as they do now, then change the set meant for PS 1.6 to an PS 1.7 compatible one at template fetch time. Perhaps with a wrapper around Smarty. However, this duplicates the work for PS 1.7 themes.

----
As always, details can be found in code comments and in commit descriptions.


benzon:
It's a good idea, but i thinking if it would be possible.

Reason: the modules themes comes with need to be supportet to then.

The reason it's a very good idea a lot of the theme dev's have added a lot of nice features in to there 1.7 releases that's not there in the 1.6.x ones.

firstred:
Just closed this as it can't be merged in the near future. Feel free to reopen when it's ready.

PR #249, opened 2017-05-29
Rewrite rules

Traumflug:
Solution for https://forum.thirtybees.com/topic/320/problem-saving-seo-url-routes


firstred:
I wonder when that mistake happened. Thanks for getting to the bottom of this and fixing it!

PR #250, opened 2017-05-30
Allow '*automatic*' PS_SHOP_DOMAIN.

Traumflug:
Let's start with something simple. Will post an RFC in the forum in a minute.

In PS 1.7 this feature worked mostly, too. Except for uploading product images (something like 'image not found', IIRC). Can't reproduce this in 30bz, product images upload and get used just fine.

Commit message:

One of the obstacles against automatic shop synchonisation is the
shop's domain name stored in the database. After copying a shop
from here to there one has to adjust these two values manually
from the admin panel, which defeats the 'automatic' part and
costs additional downtime.

Because there might be use-cases for this storage even in monoshop
mode, this storage wasn't removed. Instead, a keyword '*automatic*'
was made valid which causes ignorance of these stored values. This
way one gets the best of both, allowing an arbitrary shop domain
for those who want it and insistence on the stored value for those
who need it.

This feature works even in multishop mode. It makes sense to set
the default shop to *automatic*, then all unknown domains will
load that default shop without redirection. Known domains load
their related shop.

### DIFF
--- a/classes/Configuration.php
+++ b/classes/Configuration.php
@@ -435,6 +435,14 @@ class ConfigurationCore extends ObjectModel
                 static::$_cache[static::$definition['table']][$lang]['global'][$row['name']] = $row['value'];
             }
         }
+
+        // Adjust automatic values.
+        if (static::get('PS_SHOP_DOMAIN') === '*automatic*') {
+            static::set('PS_SHOP_DOMAIN', $_SERVER['HTTP_HOST']);
+        }
+        if (static::get('PS_SHOP_DOMAIN_SSL') === '*automatic*') {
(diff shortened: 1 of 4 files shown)


Traumflug:
Forum post: https://forum.thirtybees.com/topic/349/request-for-comment-automatic-domain-feature

Traumflug:
Two positives, one neutral comment, no objections in the forum. I think it's safe to assume this feature can go in.

firstred:
Yup. Can't wait! Thank you!

jon88seven:
Hmm i got 500 server error
[core:alert] /var/www/vhosts/domain.com/httpdocs/.htaccess: RewriteCond: cannot compile regular expression '^*automatic*$'

PR #280, opened 2017-06-30
Fix broken link to tb forum

yaniv14:


firstred:
Some well-spotted bug fixes. Thank you!

PR #284, opened 2017-07-01
Minor DbQuery cleanups.

yaniv14:
@firstred Please check before merge.


firstred:
Looks good to me. This also seems to fix some issues I'm currently having with generating invoices. Thanks!

Issue #285, opened 2017-07-01
Fix CSV tag import


Traumflug:
Pushed branch [*csv-tag-import*](https://github.com/thirtybees/thirtybees/commits/csv-tag-import) for review.

1. CSV import: don't make tags multi-language.
   CSV imports are declared as an import for one language, so import that. Before, importing tags for another language would overwrite previously imported tags.
2. CSV import: align default field separators with RFC 4180.
   This is, comma (',') as standard field separator.
   Definition of multiple values inside fields isn't standardized.
   Some applications use a semicolon (';'), others a quoted string with commas inside. See https://www.ietf.org/rfc/rfc4180.txt
3. CSV import: separate tags according to RFC 4180.
   This is done by introducing a separation function, so it can get applied to other areas as well.
4. CSV import: simplify adding an array of tags.
   There is no point in assembling a string, Tags::addTags happily accepts an array directly.
   Also don't break if one language fails, the other ones may succeed.
5. TagsCore: simplify splitting a provided string.
   There is no point in filtering, validation happens anyways.

While all of them target CSV imports, I hope they don't break other imports.

Traumflug:
Replaced by pull request #286.

PR #286, opened 2017-07-01
Fix CSV tag import

Traumflug:
This replaces issue #285. It just occured to me that one can make a PR from within the repo, too, and that this is probably better for review.

1. CSV import: don't make tags multi-language.
   CSV imports are declared as an import for one language, so import that. Before, importing tags for another language would overwrite previously imported tags.
2. CSV import: align default field separators with RFC 4180.
   This is, comma (',') as standard field separator.
   Definition of multiple values inside fields isn't standardized.
   Some applications use a semicolon (';'), others a quoted string with commas inside. See https://www.ietf.org/rfc/rfc4180.txt
3. CSV import: separate tags according to RFC 4180.
   This is done by introducing a separation function, so it can get applied to other areas as well.
4. CSV import: simplify adding an array of tags.
   There is no point in assembling a string, Tags::addTags happily accepts an array directly.
   Also don't break if one language fails, the other ones may succeed.
5. TagsCore: simplify splitting a provided string.
   There is no point in filtering, validation happens anyways.

While all of them target CSV imports, I hope they don't break other imports.

### DIFF
--- a/classes/Tag.php
+++ b/classes/Tag.php
@@ -99,7 +99,8 @@ class TagCore extends ObjectModel
      * @param int          $idLang    Language id
      * @param int          $idProduct Product id to link tags with
      * @param string|array $tagList   List of tags, as array or as a string with comas
-     * @param string       $separator
+     * @param string       $separator Separator to split a given string inot an array.
+     *                                Not needed if $tagList is an array already.
      *
      * @return bool Operation success
      *
@@ -113,7 +114,7 @@ class TagCore extends ObjectModel
         }
 
         if (!is_array($tagList)) {
-            $tagList = array_filter(array_unique(array_map('trim', preg_split('#\\'.$separator.'#', $tagList, null, PREG_SPLIT_NO_EMPTY))));
+            $tagList = explode($separator, $tagList);
         }
 
         $list = [];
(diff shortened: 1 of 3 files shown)


firstred:
Going to find out if I can merge this. Not having a comma by default also seems to mess with the Excel converter.

classes/Tag.php
```diff
@@ -113,7 +114,7 @@ public static function addTags($idLang, $idProduct, $tagList, $separator = ',')
         }
 
         if (!is_array($tagList)) {
-            $tagList = array_filter(array_unique(array_map('trim', preg_split('#\\'.$separator.'#', $tagList, null, PREG_SPLIT_NO_EMPTY))));
+            $tagList = explode($separator, $tagList);
```
firstred: shouldn't this be `fieldExplode`?

alwayspaws:
Thanks very much for looking into this. I wait before creating any new spreadsheets

firstred:
Wow, this really improved a lot. The non-CSV imports improve as well, because we can now standardize those to a `,` delimiter.

alwayspaws:
Fabulous! Can't wait for  the delimiters to be standardized. Life will be a lot easier.

firstred:
Yeah, this is how I successfully imported the tags:
exampleods

alwayspaws:
I somehow missed this post.  Thanks!

Issue #294, opened 2017-07-04
Cannot install, bankwire module failed


skoro:
Installation of ThirtyBees (1.0.x branch) failed on **Install modules** step with message:
```
An error occurred during installation...

You can use the links on the left column to go back to the previous steps, or restart the installation process by clicking here.

    Cannot install module "bankwire"
```
Apache log is empty, app log contains the same message: `Cannot install module "bankwire"`.

Details:
PHP: PHP 7.1.6-2~ubuntu14.04.1+deb.sury.org+1
Web server: Apache/2.4.7 (Ubuntu)
MySQL: 5.6.33-0ubuntu0.14.04.1

firstred:
Hey,

I knew that happened a lot yesterday, but are you still seeing it at the moment?

skoro:
Yes, today I have cloned the repo and tried to install.

firstred:
I have installed thirty bees 1.0.2 on a multitude of servers now and haven't seen the error once. My guess is you might locally still have some old files, because the build used to break exactly like that until yesterday. It's too much of a coincidence imho. Could you try again with a totally fresh clone? Or dig your logs and share the error? Thanks!

skoro:
Hi,
You can reproduce this bug if you during installation select Ukraine country. I have found that **bankwire** module failed in `PaymentModule::addCheckboxCurrencyRestrictionsForModule()` method:
```
            if (!Db::getInstance()->insert(
                'module_currency',
                [
                    'id_module' => (int) $this->id,
                    'id_shop' => (int) $s,
                    'id_currency' => ['type' => 'sql', 'value' => '(SELECT `id_currency` FROM `'._DB_PREFIX_.'currency` WHERE `deleted` = 0)'],
                ]
            )) {
                return false;
            }
```

This insert cannot be executed with following data in **currency** table:
```
+-------------+--------------+----------+--------------+------+-------+--------+----------+-----------------+---------+--------+
| id_currency | name         | iso_code | iso_code_num | sign | blank | format | decimals | conversion_rate | deleted | active |
+-------------+--------------+----------+--------------+------+-------+--------+----------+-----------------+---------+--------+
|           1 | Гривна       | UAH      | 980          | ₴    |     1 |      2 |        1 |        1.000000 |       0 |      1 |
|           2 | Евро         | EUR      | 978          | €    |     1 |      2 |        1 |        1.000000 |       0 |      1 |
[...]

PR #296, opened 2017-07-05
Fix bankwire module installation for countries with multicurrencies.

skoro:
This pull request resolves the issue https://github.com/thirtybees/thirtybees/issues/294


firstred:
Thank you!

PR #307, opened 2017-07-12
Installreports

Traumflug:
Today I had enough of these vague "directory not recursive writable" messages at installation time, so I digged into this.

- Error messages created during these tests get now forwarded to the installer screen. With this, finding the culprit was easy: a file inside a `.git` folder.
- Also rewrote this recursive test, it now tests directories, too. RecursiveDirectoryIterator lists only plain files.
- And excluded `.git` directories, of course.

Not pushed to *1.0.x* directly, because I'm not sure whether it's a good idea to do changes in such a sensible area shortly before a release.

### DIFF
--- a/classes/ConfigurationTest.php
+++ b/classes/ConfigurationTest.php
@@ -415,15 +415,22 @@ class ConfigurationTestCore
         }
 
         if ($recursive) {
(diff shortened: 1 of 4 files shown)


firstred:
Makes sense! Thanks!

PR #311, opened 2017-07-15
Fix custom code CSS and METAS not output to webpage due to use of inc…

roband7:
…orrect configuration key.


firstred:
Thanks, that's what we have been looking for!

PR #313, opened 2017-07-18
fixed employee password email template.

yaniv14:


admin-dev/themes/default/sass/partials/_rtl.sass
```diff
@@ -152,30 +152,28 @@ html
...
+					border-right: 14px solid $main-color
+					border-left: 0 !important
+			&.done
+				.chevron:after
+					border-right: 14px solid $brand-success
+					border-left: 0 !important
+	.actionBar
+		.msgBox
 			float: right !important
-	#carrier_logo_block
```
firstred: Why is this one removed?

admin-dev/themes/default/sass/partials/_rtl.sass
```diff
@@ -152,30 +152,28 @@ html
...
-		left: 10px
-		right: auto !important
-	.actionBar a
```
firstred: Does the actionbar link style have to be removed as well?

yaniv14: carrier_logo_block was my mistake (no need for that).
the actionbar moved above (better)

firstred: Alrighty, makes sense

firstred:
Thanks!

PR #322, opened 2017-07-24
Fix DB::getInstance()->NumRows() deprecated or not working anymore

Zengraph:
the Backup system created archives files but empty of datas, only tables structure were saved.

http://doc.prestashop.com/display/PS15/DB+class+best+practices#DBclassbestpractices-NumRows()

says it's deprecated.. and the fact is that it was not working anymore.. it returned NULL, so the test after was failed and no data was sent.. Backup were only including structure no datas, and now it's fixed !


firstred:
Going for the other solution: https://github.com/thirtybees/thirtybees/commit/c1a647ff3902565adfb235788bd1360adc3c5ef8

But thanks nonetheless!

Issue #324, opened 2017-07-28
TB acts as if multilanguage is always active due to regression in countActiveLanguages


roband7:
This is a regression in 1.0.2, see discussion on forum:
https://forum.thirtybees.com/topic/608/iso-code-added-to-url-after-migration/

It´s caused by the change to classes/Language.php function countActiveLanguages() in commit ID 5b225bb on the 22´nd of June.

The clause "where active = 1" was removed by mistake.

PR #325, opened 2017-07-29
fix countActiveLanguages() DbQuery, was missing `active` check

yaniv14:
reported: https://github.com/thirtybees/thirtybees/issues/324


firstred:
Thanks!

Issue #326, opened 2017-07-30
Tax & shipping fee calculation bug in AEUC


30knees:
As discussed [here](https://forum.thirtybees.com/topic/584/bug-in-eu-advanced-compliance-with-shipping-taxes), there's a bug in the AEUC. 

**Problem**

- tb takes the shipping price as including taxes.
- AEUC takes the shipping price as excluding taxes.
- Because of this, when “Proportionate tax for shipping and wrapping” is enabled for AEUC, the actual shipping fee is calculated dynamically based on the proportionate tax rate of the goods in the cart. This means that the final shipping fee cannot be stated in advance to the customer. This is both inconvenient for the merchant and the customer and in violation of EU law, which requires that the shipping price needs to be stated in advance.

**Solution**
- AEUC should take the shipping price as including taxes.

Nobodaddy:
You can find the bug in function *getPackageShippingCost* of class **Cart.php**
It starts absolutely right line 1882 

           if (Configuration::get('PS_ATCP_SHIPWRAP')) {
                // With PS_ATCP_SHIPWRAP, pre-tax price is deduced
                // from post tax price, so no $carrier_tax here
                // even though it sounds weird.
                $carrierTax = 0;
                (...)

But it ends wrong from line 2003 on, because because the avaerage tax is **added** to the shipping costs and not  **deducted** from this figure which would be correct:

        if (Configuration::get('PS_ATCP_SHIPWRAP')) {
            if ($useTax) {
                // With PS_ATCP_SHIPWRAP, we apply the proportionate tax rate to the shipping
                // costs. This is on purpose and required in many countries in the European Union.
                $shippingCost *= (1 + $this->getAverageProductsTaxRate());
            }

Nobodaddy:
Solution: [https://github.com/thirtybees/thirtybees/pull/327](https://github.com/thirtybees/thirtybees/pull/327)

doclucas:
Hasn't this been fixed with this?
https://github.com/thirtybees/thirtybees/pull/327
https://github.com/thirtybees/thirtybees/pull/327/commits/cd8a332597a9aea5d918f969ac337c0fa37e35e6

mke-tb:
@30knees who will take a look and let us know whether it is still reproducible. Thanks!

30knees:
Sorry, this totally slipped me. 

I'm running the latest bleeding edge 3.1.3 of AEUC.

[...]

PR #327, opened 2017-07-30
CO: Bugfix Cart.php shipping fee with AEUC enabled

Nobodaddy:
thirty bees ignores that with AEUC enabled the shipping fee is treated as tax **included**. This is necessary to fulfil legal requirements in Europe. The customer has to be informed about the full shipping costs **before** he places the order.

### DIFF
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -2000,13 +2000,7 @@ class CartCore extends ObjectModel
             }
         }
 
-        if (Configuration::get('PS_ATCP_SHIPWRAP')) {
-            if ($useTax) {
-                // With PS_ATCP_SHIPWRAP, we apply the proportionate tax rate to the shipping
-                // costs. This is on purpose and required in many countries in the European Union.
-                $shippingCost *= (1 + $this->getAverageProductsTaxRate());
-            }
-        } else {
+        if (!Configuration::get('PS_ATCP_SHIPWRAP')) {
             // Apply tax
             if ($useTax && isset($carrierTax)) {
                 $shippingCost *= 1 + ($carrierTax / 100);


firstred:
I'm not sure if that fixes all. There are a lot of cases that are still funky. What happens with this change it that tax is subtracted from the tax excl. field, which is more like a hack to accomplish what we want. It seems like there indeed is no other way at the moment, but by releasing `1.0.2` we pretty much "locked" the field's functionality. I would like to recommend to add a tax incl/excl selector to the Advanced EU compliance module and read this setting when calculating taxes. This way we can probably inject the right amount in the tax excl fields before it hits the Cart object.

Nobodaddy:
Nothing happens! 😊
I just deactivated the **normal** behavior for shipping tax in case the option *proportionate tax on shipping and handling* in AEUC is enabled. That's all!

30knees:
> Nothing happens! 😊
> I just deactivated the **standard** behavior for shipping tax in case the option _proportionate tax on shipping and handling_ in AEUC is enabled. That's all!
> And you need to keep this treatment of shipping costs (input is treated as tax incl.) alive because otherwise you would cause a major issue in AEUC.

How do you use it? For me it doesn't seem to be working, see eg:

https://github.com/thirtybees/thirtybees/issues/326#issuecomment-730280442

30knees:
@Nobodaddy I implemented your fix and while it works to a certain extent, something isn't right:

image

1. The base prices for shipping are correct **including** taxes. But they should be excluding taxes.
2. The column "total taxes" for shipping is incorrect.

Traumflug:
No surprise that the fix above doesn't work. It removes this kind of tax calculation, accordingly the total of taxes on shipping is always zero. Which isn't the intended outcome.

PR #329, opened 2017-07-31
CO: Bugfix product total amount tax incl/excl for option ROUND_LINE

Nobodaddy:
With option *ROUND_LINE* the total price in now correct when there's more than one product ordered. But still there's no algorithm for the option ROUND_TOTAL.

### DIFF
--- a/classes/pdf/HTMLTemplateInvoice.php
+++ b/classes/pdf/HTMLTemplateInvoice.php
@@ -153,7 +153,7 @@ class HTMLTemplateInvoiceCore extends HTMLTemplate
                     $orderDetail['total_price_tax_excl_including_ecotax'] = $orderDetail['unit_price_tax_excl_including_ecotax'] * $orderDetail['product_quantity'];
                     break;
                 case Order::ROUND_LINE:
-                    $orderDetail['total_price_tax_excl'] = Tools::ps_round(($orderDetail['total_price_tax_excl'] * $orderDetail['product_quantity']), _PS_PRICE_DISPLAY_PRECISION_);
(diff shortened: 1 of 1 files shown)


firstred:
Have you tested both types (with and without eco tax)?

I believe this fix should only be applied to the eco tax one.

Nobodaddy:
Nope, we do not have ecotax in Germany, therefor it wasn't activated.

classes/pdf/HTMLTemplateInvoice.php
```diff
@@ -153,8 +153,8 @@ public function getContent()
                     $orderDetail['total_price_tax_excl_including_ecotax'] = $orderDetail['unit_price_tax_excl_including_ecotax'] * $orderDetail['product_quantity'];
                     break;
                 case Order::ROUND_LINE:
-                    $orderDetail['total_price_tax_excl'] = Tools::ps_round($orderDetail['total_price_tax_excl'], _PS_PRICE_DISPLAY_PRECISION_);
-                    $orderDetail['total_price_tax_excl_including_ecotax'] = Tools::ps_round($orderDetail['unit_price_tax_excl_including_ecotax'], _PS_PRICE_DISPLAY_PRECISION_);
+                    $orderDetail['total_price_tax_excl'] = Tools::ps_round(($orderDetail['total_price_tax_excl'] * $orderDetail['product_quantity']), _PS_PRICE_DISPLAY_PRECISION_);
```
firstred: I believe it's okay with the quantity removed from this line. `total_price_tax_excl` should always be the line total and therefore does not need a quantity.

firstred:
Thanks for giving permission to edit the branch. I have made the changes and am going to merge this pull request. Let's see if that fixes the issue!

PR #339, opened 2017-08-13
Unit tests for the Validate class

kublaios:
Tests for each function in the `Validate` class, there are 5 functions that didn't pass:  
  
* `isOrderWay`
* `isWeightUnit`
* `isDistanceUnit`
* `isLabel`
* `isOrderInvoiceNumber`  
  
Notice that they've marked with `TODO`s for the time being and returning `true` by default to pass the tests. Their tests are placeholders until solved.


roband7:
I´m happy and impressed with the work you´re doing here!

In my opinion the lack of test coverage was always a major flaw in Prestashop, and one of the reasons quality always took a dive after major changes.

That thirty bees does things better again is great.

Keep up the good work!

PR #348, opened 2017-08-28
Fix import products without Force all ID numbers.

firstred:
Importing products without option Force all ID numbers but with
the ID column provided resulted into updating the existing
products.

Discussion see
https://www.prestashop.com/forums/topic/566115-prestashop-17-is-now-available/?p=2596950
https://forum.thirtybees.com/topic/731/error-importing-csv-categories-file/14

Original patch by Gabriel Arama <arama.gabriel@gmail.com>, made
for PrestaShop 1.7.


firstred:
Looks good to me. Thanks!

PR #350, opened 2017-08-31
Fixed bug: unable to remove product from pack

yaniv14:
reported in https://forum.thirtybees.com/topic/807/can-t-delete-a-product-in-a-pack


firstred:
Yes, confirmed to be working here as well. Merged, thanks!

PR #356, opened 2017-09-11
Dev

Lathanao:
Hi,

I just fix createDatabase() in ObectModel.php for an unfail installation of the blog module.
I tried installing some others modules with this fix, everything looks well.
(The modifications I added in the beesblog repostory are useness now but i will try continue to improve this blog module)

Regards.

### DIFF
--- a/classes/ObjectModel.php
+++ b/classes/ObjectModel.php
@@ -2197,7 +2197,6 @@ abstract class ObjectModelCore implements Core_Foundation_Database_EntityInterfa
         }
 
         $definition = static::getDefinition($className);
-
         $sql = 'CREATE TABLE IF NOT EXISTS `'._DB_PREFIX_.bqSQL($definition['table']).'` (';
         $sql .= '`'.$definition['primary'].'` INTEGER NOT NULL PRIMARY KEY AUTO_INCREMENT,';
         foreach ($definition['fields'] as $fieldName => $field) {
@@ -2218,8 +2217,6 @@ abstract class ObjectModelCore implements Core_Foundation_Database_EntityInterfa
         }
         $sql = trim($sql, ',');
         $sql .= ')';
-
-Tools::p($sql);
 
         try {
             $success &= Db::getInstance()->execute($sql);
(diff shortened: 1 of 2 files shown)


Traumflug:
Thank you for this contribution!

It'd be a good idea to remove the debugging code from these commits. Like doing a `git rebase -i 1.0.x` and editing them in-place or squashing the last commit with the second one. This PR will change accordingly if you `push -f` the new set of commits.

Can't tell whether these SQL changes are a good idea.

firstred:
Thanks for the suggestions! You are probably using MySQL 5.7 as some of the stuff is deprecated on that version and breaks it indeed. I'll do a review in a sec. :)

classes/ObjectModel.php
```diff
@@ -2206,7 +2206,28 @@ public static function createDatabase($className = null)
             if (isset($field['lang']) && $field['lang'] || isset($field['shop']) && $field['shop']) {
                 continue;
             }
-            $sql .= '`'.$fieldName.'` '.$field['db_type'];
+
```
firstred: Can you go through the switch in case `$field['db_type']` is missing? It would otherwise limit the possibilities that modules do have right now.

firstred: It is a great idea to add some defaults. Less typing to add the default types. Always nice :)

classes/ObjectModel.php
```diff
@@ -2245,10 +2266,10 @@ public static function createDatabase($className = null)
             }
 
             // Lang field
-            $sql .= '`id_lang` INT(11) DEFAULT NULL,';
+            $sql .= '`id_lang` INT(11) NOT NULL,';
```
firstred: Great catch! Makes no sense to have a lang table without the `id_lang` filled!

firstred:
Thanks again for this contribution. I am just checking this again and noticed the last things have not been changed :(
I am still going to merge it and change those items myself, making sure this lands in the next version of thirty bees. It is a great change and we thank you for that!

PR #363, opened 2017-09-21
Tax on invoice is broken

eschiendorfer:
I readded the old value from PS. The new line from thirtybees breaks the taxe rates on pdf invoice. Instead of 2.500% it just shows 2%...
No idea why, but on my template it acts like this...
Not even $rate = $row['tax_rate']; is working

### DIFF
--- a/classes/order/OrderInvoice.php
+++ b/classes/order/OrderInvoice.php
@@ -545,7 +545,7 @@ class OrderInvoiceCore extends ObjectModel
         }
 
         foreach ($details as $row) {
-            $rate = (float) round($row['tax_rate'], 3);
+            $rate = sprintf('%.3f', $row['tax_rate']);
             if (!isset($breakdown[$rate])) {
                 $breakdown[$rate] = [
                     'total_price_tax_excl' => 0,


firstred:
Strange, a `round(X, 3);` does not round a `2.500` to `2`. Can you share what's in your invoice template?

eschiendorfer:
`{foreach $product_tax_breakdown as $rate => $product_tax_infos}
<tr style="line-height:16px;background-color:{cycle values='#FFF,#FFF'};">
    <td style="width: 25%">
        {if !isset($pdf_product_tax_written)}
        {l s='Products' pdf='true'}
        {assign var=pdf_product_tax_written value=1}
        {/if}
    </td>
    <td style="width: 20%; text-align: right;">{$rate} %</td>
    {if !$use_one_after_another_method}
    <td style="width: 25%; text-align: right;">
        {if isset($is_order_slip) && $is_order_slip}- {/if}{displayPrice currency=$order->id_currency price=$product_tax_infos.total_price_tax_excl}
    </td>
    {/if}
    <td style="width: 20%; text-align: right;">{if isset($is_order_slip) && $is_order_slip}- {/if}{displayPrice currency=$order->id_currency price=$product_tax_infos.total_amount}</td>
</tr>
{/foreach}`

firstred:
Can you try to align that with tb default templates? I believe you have to force it to display as a float
Example:
```
<td class="right white">
	{$line.rate|floatval} %
</td>
```

eschiendorfer:
I updated now my whole invoice.tax-tab.tpl file. Now it's working without the Core Change!

PR #370, opened 2017-10-05
Resolves Bug for filtering "name" column

eschiendorfer:
The column name has a Join as 'b' not 'pl'. That's why there is an error when trying to filter the column.


Traumflug:
You know how these `filter_key`s work? Congratulations! I don't :-)

As you apparently know what you're doing I moved the commit to *1.0.x*. Thank you very much!

eschiendorfer:
Well to be honest. I am new into this things. But I guess the filter_key works just like in SQL. filter_key=>'b!name' just adds something to the query like this: " WHERE b.name LIKE '%input%'

alwayspaws:
Emanuel, you are more knowledgeable than you realize. Nice catch!

PR #380, opened 2017-10-15
add table-responsive class to admin css

yaniv14:
reported @ https://forum.thirtybees.com/topic/943/small-payment-css-overlap-on-small-screens


firstred:
Nice one! Thanks.

PR #381, opened 2017-10-16
update do.xml to correct info

RaptorX:
My country information has been always setup incorrectly, this should be the default information with the latest country info.


Traumflug:
Thank you very much for your contribution! I fixed whitespace and applied the commit.

PR #385, opened 2017-10-25
Fix SQL export separator to default comma

RaptorX:


firstred:
And break it for everyone else? No way!
I'm sorry but we cannot change this just because you prefer a comma. Why not add an option to the admin page, so people can choose?

RaptorX:
The reason why i thought of changing it is because I saw that in the CSV import section **you guys changed it to comma as the default**, it was only logical to think that was the intention.

How about this: make it optional and not hadcoded... because as it stands then I have to replace all ';' whenever I export and I asume a lot of non europeans who might use the products too...

Is not that I prefer it hahaha, thats the whole name of the thing... **Comma** Separated Values file... actually Prestashop developers changed to semicolon because **they prefered it that way** but thats not the standard :P

But I get your point...

firstred:
Yeah, I can imagine it's not the best if you need a different separator. But do realize that you will have to do the following first to make it work:
- Escape values that contain the new separator (probably with quotes `"`)
- Add a new input to the page for the separator, like on the CSV Import page
- Grab this value and use it as the new separator

I think it can be done and we would definitely accept a change like that, but it has to default to the current situation and the new separator has to be configurable. Would you be willing to add these features? We'll merge it when everything is complete.

RaptorX:
Well yes, lets see how far I can take it :)

firstred:
Alrighty, thanks for your efforts so far!

Traumflug:
> you guys changed it to comma as the default

Yes, this was me, for the reasons you described. It's CSV, not SSV.

During refining this area a bit I certainly had the wish to rewrite more than I did. But changing methods changes the API for modules, so a no-go for 1.0.x.

RaptorX:
>  But changing methods changes the API for modules, so a no-go for 1.0.x.

I think this is the point I was missing when I submitted the pull request. I will take that in consideration in the future.

Is my first time contributing so yeah... will be more careful ^_^
Is there a place you guys hang around so that I can ask for feedback before submitting a pull request (for example talking about ideas I might have that you guys are already doing or that might not be a good idea to code right now)

Traumflug:
No need to worry, you PR was the right thing to do. There is a reason why there's a discussion section for each PR :-)

PR #399, opened 2017-11-22
Removed hardcoded timeout in AdminController.php

benzon:
Removed the hardcoded 900 sec replaced with ((int)Configuration::get('PS_COOKIE_LIFETIME_BO') * 3600) - so now it's using the lifetime set in BO.

### DIFF
--- a/classes/controller/AdminController.php
+++ b/classes/controller/AdminController.php
@@ -3839,7 +3839,7 @@ class AdminControllerCore extends Controller
         }
 
         if (isset(Context::getContext()->cookie->last_activity)) {
-            if ($this->context->cookie->last_activity + $cookie_lifetime < time()) {
+            if ($this->context->cookie->last_activity + ((int)Configuration::get('PS_COOKIE_LIFETIME_BO') * 3600) < time()) {
                 $this->context->employee->logout();
             } else {
                 $this->context->cookie->last_activity = time();
--- a/classes/controller/AdminController.php
+++ b/classes/controller/AdminController.php
@@ -3839,7 +3839,7 @@ class AdminControllerCore extends Controller
         }
 
         if (isset(Context::getContext()->cookie->last_activity)) {
-            if ($this->context->cookie->last_activity + 900 < time()) {
+            if ($this->context->cookie->last_activity + $cookie_lifetime < time()) {
                 $this->context->employee->logout();
             } else {
                 $this->context->cookie->last_activity = time();


benzon:
Should help with a long standing issue from PS, where BO users get's logged out if they are not active on the page for 15 min, even if Stay logged in was checked.

doekia:
This patch renders the "Stay logged" option useless. If now I want not to have a persistant session, how do I do it? 

The PS_COOKIE_LIFETIME_BO on vanilla PS obey the "Stay logged" is checked. AdminLoginController does not inject the last_activity time in the cookie hence you are not logged out unles the cookie it self expires

benzon:
It dos actually not obey at all doekia that is wrong.

The 900 is hardcoded for all logins, there is 0 check in that part for the sstay logged in.

benzon:
then the function needs a total rewright where it takes in to account if Stay Logged In is checked.

But it actually dos not work we have had a lot of issues with some of our employes getting logged out, if they are inactive for 15 min, even tho stay logged in is checked.

benzon:
And if you read in BO - the Lifetime of back office cookies is Set the amount of hours during which the back office cookies are valid. After that amount of time, the thirty bees user will have to log in again, so this is the correct action insted of the hardcorded 900 sec

Traumflug:
First, thanks for pointing out the issue. If users get logged out unexpectedly, something is going wrong.

Looking at AdminLoginController.php line 265 I can see this:
```php
if (!Tools::getValue('stay_logged_in')) {
    $cookie->last_activity = time();
}
```
This is, as far as my grep'ing goes, the only place where this "Stay logged in" checkbox is handled in PHP code. It means that `$cookie->last_activity` isn't even set for employees wanting to be logged in permanently. It also means that the code you modified shouldn't even be reached, see the `if (isset(...))` around it.

It looks like this 'last_activity' is set when it shouldn't be. If you have a reproducible test case, @benzon, please investigate a bit in this direction, whether this guess is true. I'll try a few ideas, too.

doekia:
No, rewrite of code needed:
classes/controller/AdminController.php
if (isset(Context::getContext()->cookie->last_activity)) {
           if ($this->context->cookie->last_activity + 900 < time()) {
                 $this->context->employee->logout();

controllers/admin/AdminLoginController.php
if (!Tools::getValue('stay_logged_in')) {
        $cookie->last_activity = time();
}

benzon:
I can do a video of it from different PC's it's a know issue.

https://www.prestashop.com/forums/topic/382570-keep-getting-logged-out-of-admin-panel/

https://www.prestashop.com/forums/topic/342546-solved-bo-login-session/ <-- This is actually what the fix is based on.

benzon:
But agree this is not a perfect fix, the perfect fix would be to fix the stay_logged_in feature that's is not saving the value correct.

Traumflug:
Before doing hacks one should always understand what's going on. Your PR essentially disables the "Stay logged in" unchecked feature. Having it disabled, one could entirely remove the checkbox entirely.

With "Stay logged in" checked, this `stay_logged_in` value doesn't even exist. Well, it shouldn't. I tried around whatever I could imagine, but this value appears with this box unchecked, only. So I wonder why it exists at your installation.

Could you add two lines of code in AdminController? To make it look like this:

```php
file_put_contents(_PS_ROOT_DIR_.'/config/debug', "\n".'isset(last_activity) '.var_export(isset($this->context->cookie->last_activity), true)."\n", FILE_APPEND);
file_put_contents(_PS_ROOT_DIR_.'/config/debug', 'last_activity '.var_export($this->context->cookie->last_activity, true)."\n", FILE_APPEND);
        if (isset(Context::getContext()->cookie->last_activity)) {
            if ($this->context->cookie->last_activity + 900 < time()) {
                $this->context->employee->logout();
            } else {
                $this->context->cookie->last_activity = time();
            }
        }
```
This won't affect operations, but create a file `config/debug`. Each page request adds two lines in this file and all lines should report 'false' with "Stay logged in" checked. You can also change this `900` to `20` to reproduce the issue faster.

That said, I wouldn't be opposed to removing this checkbox entirely. It's checked by default and it's hardly imaginable employees willingly uncheck it.

guestisp:
Isn't last_activity only set on PHP side?
Thus, it's normal that without activity employee get disconnected, as PHP is only involve during a page load

To keep cookie alive we need an ajax keepalive called automatically with setInterval from a JavaScript

doekia:
A lots of mis-comprehension here:

When you are at BO login (AdminLoginController), a cookie is created with your employee with expiration set at now() + (PS_COOKIE_LIFETIME_BO*3600),  thru config/config.inc.php detecting PS_ADMIN_DIR.

Should you NOT check the "Stay logged"  the cookie is added a last_activity thru AdminLoginController.
Every time your browser request a BO page, if and only if this last_activity is present it is compared to 900s. If more than 900s you are logout() else the last_activity is updated.

If during BO login you have checked the "Stay logged", no property last_activity exists in the cookie, hence no 900s check is perform. Your browser will send the cookie until it expire.

Bear in mind that each time you request à page the cookie get its expiration date rewritten.

-- 

Be sure you have one single page open while doing your tests, and that you are thruly logout() by this line of code. IPCHECK could be another source of annoying logout(). Another source of problem is FO navigation while been BO connected and the ******* module cronjobs that fake PS_ADMIN_DIR and create all sorts of annoyance.

benzon:
@doekia IP should not be an issue at all in my case i got a static ip assigned.

The think is here, if i use the fix i made it actually works, i dont get logged out any longer, when i check keep logged in, if the fix is removed 15 min goes and im logged out.

benzon:
@Traumflug @doekia 

Figured it out, the issue can be reproduced the following way.

Login with out checking keep logged in, close your browser, go back to admin it will ask you to loggin again, i got a small script to view the cookies with out logging in and it clearly shows [last_activity] => 1511423649 - is still set, so when you login even tho you check the Keep Logged In the last_activity is still set, so you actually end up in the logout loop.

doekia:
Having a static IP does not mean your BO sees the same IP. This for instance is not the case when using some proxy upfront that is not quite well configured (OVH is notoriously known to have such issue with either their "invisible varnish" (.ovhconfig => environment=production) and they ssl front end.

But if your patch makes it to work for you, I suspect cronjobs (even uninstalled since it is a need instance module) or other component that corrupt your cookie (i.e: injects last_activity).

doekia:
Ok I propose a patch even do I have never quite faced your situation:

In AdminLoginController.php:
if (!Tools::getValue('stay_logged_in')) {
                    $cookie->last_activity = time();
}
else {
  unset($cookie->last_activity);
}

benzon:
@doekia i know my hosting setup 100% since im quite well versed when it comes to that, and there is nothing there that can interfere with my IP :)

benzon:
But think in general a new field should be added to Admin BO, allowing to configure the timeout insted of the hardcoded 900, and the text of the current field should be changed, since it's wrong currently, it shold hold a note of being the length of cookie for stay logged in.

And then there should be a new field called Login Timeout, where the 900 sec can be configured, this would be the right way to do it, and everything else i have used, plesk cpanel you name it, got a option for this.

@firstred would this be possible with out breaking things adding this ?

doekia:
There is such a field, it is PS_COOKIE_LIFETIME_BO. When I request not to be remembered, the session expires after 900s of inactivity. Which is 15mn (for non persistent session) and PS_COOKIE_LIFETIME_BO * 3600 for persistent. 

I'm pretty sure that the issues occur because of "old cookie". My proposed patch should fix this while keeping the code operative.

benzon:
@doekia that i know,.

Stil think it's an old design flaw from PS that it's not possible to configure the non persistent session timeout.

If you look at all other major software, ESXI Web UI, Plesk, Cpanel, even thing Magento got the option to configure the timeout, and it would make sence to allow this to be configured, so if you want a temp login but want it to be 60 min, you can configure this in BO.

I am currently testing the other solution to see if this fixes the issue or not.

Traumflug:
> Login with out checking keep logged in, close your browser, go back to admin it will ask you to loggin again

Thank you very much, @benzon, this is helpful. I still can't reproduce it in either Firefox or Chrome, because closing the browser doesn't log me out. After restart I see the dashboard immediately, without logging in. Same for "Stay logged in" checked or unchecked.

Nevertheless this description is detailed enough to see what's going on, apparently your browser tries to "enhance" something in unexpected ways. This additional line @doekia described looks like the right thing to do. Will make a commit in a minute.

Regarding these 900 seconds: I don't think they need to be user settable. They mean "log out immediately after being done". It just happens that detecting "being done" isn't easy, software can't detect whether a user looks at the screen or is on its way home already. To have a user defined logout time there's Backoffice -> Administration -> Preferences -> General (which uses cookie lifetime).

benzon:
Hmm i think it renders the cookie lifetime invalid in the last end.

Since the last_active timer is set if the stay logged in is not ticked, and it can in some sence figure out if you are idle or not, the last_active timer is refreshed on every page switch.

One of my employes have had the logout issue alot, so we hardcoded a longer timeout, since there is nothing that makes you more mad then being logged out whill translating, and loosing all your work.

Traumflug:
> One of my employes have had the logout issue alot, so we hardcoded a longer timeout, since there is nothing that makes you more mad then being logged out whill translating, and loosing all your work.

This should not happen, of course. I hope this additional line removes this type of issue. As you're about to test whether the fix works, I'll wait until this is testing is done.

doekia:
The main purpose of "not remember" session is should I connect to my BO from a non-trusted network (cyber café, public wifi, ...). Till I act upon my shop the session is valid, but if all a sudden the connection drops, I don't quite care, a man in a middle will not be able to use the token/cookie after 15mn.

Problem of letting this param configurable is that poorly educated shop owner will change this value to some ridiculously high  value not knowing what it exactly does. Like when you have people doing crazy chmod 777 on their entire install. If the trend goes at having an extra tunable param for "non remembered session", I recommend the value to be caped to some reasonable timeout

benzon:
The wird thing is it's not a consistent issue, i personally got it on 1 of our 3 shops, and 2 employes got it on all 3 sites.

benzon:
i know what the point of the time out is, but again i feel that should be up to every shop owner to decide how long/short they want that timeout to be.

RaptorX:
"One of my employes have had the logout issue alot, so we hardcoded a longer timeout, since there is nothing that makes you more mad then being logged out whill translating, and loosing all your work." --

This may be irrelevant but if the browser is set to delete cookies when closing it might cause this issue, might be interesting to see why this particular employee had the issue more than others.

Dh42:
The intermittent problem could be related to ipv6 issues. I have seen that in the past when it tried to validate cookies and the ip is the dame but the ipv6 has changed.

benzon:
Oh we actually dont use IPv6 here i DK at all, i dont think there is one single internet provider yet that offers IPv6

Traumflug:
Thank you for your contributions, everybody, commit is pushed.

PR #401, opened 2017-11-24
Fixing SQL query for product attributes

Zengraph:
There was an "AND" missing in the Sql query that causes crash and error DB repport.


firstred:
Let's merge that asap. Thanks!

PR #403, opened 2017-11-29
Fixes colorOnBackground admin controllers' property not working

gonssal:
Fixes colorOnBackground admin controllers' property not working due to bootstrap setting background on TDs, thus making the TR background color hidden.


firstred:
Nice one, thanks!

The background color style indeed needs to be applied to the `<td>` rather than the row itself. Looks good to me.

PR #413, opened 2017-12-13
Fixes AdminController's processDeleteImage() redirect after URL

gonssal:
The redirect URL should be `update` instead of `add`, because you can delete the image when you are editing the object, not when adding it.


firstred:
That makes sense. Thank you!

PR #415, opened 2017-12-15
Exception handling/documentation

firstred:
Improve and document exception handling of the Classes. This will make sure we can handle them more effectively in Controllers in the future.
One example already is the importer controller. Problems like this should disappear by handling the exceptions properly: http://rameerez.com/prestashop-import-error/

### DIFF
--- a/classes/AdminTab.php
+++ b/classes/AdminTab.php
@@ -868,7 +868,6 @@ abstract class AdminTabCore
      *
      * @deprecated 1.0.0
      * @throws PrestaShopException
-     * @throws PrestaShopException
      */
     public function displayListContent($token = null)
     {
@@ -1958,7 +1957,6 @@ abstract class AdminTabCore
      * @param bool $idObject
      *
      * @deprecated 1.0.0
-     * @throws PrestaShopException
      * @throws PrestaShopException
      */
     protected function updateAssoShop($idObject = false)
--- a/classes/AdminTab.php
+++ b/classes/AdminTab.php
@@ -318,6 +318,7 @@ abstract class AdminTabCore
      * @deprecated 1.0.0
      *
      * @return bool
+     * @throws PrestaShopException
      */
     public function includeSubTab($methodname, $actions = [])
     {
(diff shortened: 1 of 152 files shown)


firstred:
Analyzing the exceptions effectively requires PhpStorm 2017.3 or higher.

classes/AdminTab.php
```diff
@@ -864,6 +867,8 @@ public function displayListHeader($token = null)
      * @param string|null $token
      *
      * @deprecated 1.0.0
+     * @throws PrestaShopException
+     * @throws PrestaShopException
```
Traumflug: Duplicate :-)

classes/AdminTab.php
```diff
@@ -1946,6 +1958,8 @@ protected function afterUpdate($object)
      * @param bool $idObject
      *
      * @deprecated 1.0.0
+     * @throws PrestaShopException
+     * @throws PrestaShopException
```
Traumflug: Also a Duplicate.

classes/AdminTab.php
firstred: Yeah, I noticed PhpStorm loves to duplicate it when my desktop hangs. Let's get that removed :)

gonssal:
Looks good, works with VSCode too. I'm not sure if @version lines are a good idea though, they are extremely tedious to maintain and, specially on big functions, will probably stay the same forever. 

I think git history is more than enough to keep track of changes.

firstred:
> Looks good, works with VSCode too. 

Nice, I am trying to optimize code for both. Didn't know VSCode already understands the exception tags. Great!

 > I'm not sure if @Version lines are a good idea though, they are extremely tedious to maintain

They are! I'm actually not even trying to keep up anymore and I don't think anyone should. The `@since` one is more important. Even though we're trying to stick to SemVer as much as possible, we sometimes end up with new functions to get tougher bugs out of the way. Even though these new methods might be public, the `@since` will warn developers to avoid these functions on earlier versions.

gonssal:
VScode needs an extension.

The `@since` ones are great and useful, the `@version` ones, not so much. I'd just get rid of them to be honest.

PR #421, opened 2017-12-23
[Core > Stock.php] Bug Fix

TobyMudito:
Remove unnecessary AND on where.


SLiCK-303:
As we are doing a check before that..
if (!(int) $idProduct && !is_int($idProductAttribute) && !(int) $idWarehouse) {
            return false;
}
 checking if $idProductAttribute is an int, we really dont need to do it again, the line should just be..
->where('`id_product_attribute` = '.(int) $idProductAttribute)

firstred:
Great! Thank you!

PR #428, opened 2018-01-01
Fix for Specific Prices not duplicating

SLiCK-303:
This fixes Specific Prices not get duplicated when you duplicate a product


SLiCK-303:
I was hoping to get these in three different pull requests, but oh well, here they are.

firstred:
Thanks! I am going to merge it before you push more unrelated commits to the same pull request :P
Make sure you start from a clean branch (`1.0.x` => `checkout -b newbranch` or something like that) with the next PR.

PR #429, opened 2018-01-02
Add APCu Smarty cache type

firstred:
On new year's day my store crashed because the Smarty cache couldn't write all the new files that had to be generated due to expired offers, fast enough to keep the store running. Using MySQL would cause too much load on our server. Using the APCu turned out to be a nice solution. Therefore I am sharing this solution.

Not sure yet if this could become part of thirty bees' core. It is not the easiest cache to configure, but once it is up and running it can give a nice performance boost.

NOTE: if you see
`PHP Warning: apc_store(): Unable to allocate memory for pool.`
it is time to increase the PHP setting `apc.shm_size`. The actual number needed depends on the store.


inetbiz:
Ansible need to be updated for this?

firstred:
Not sure, yet. My store was 4 times slower on APCu. I reverted to file system again and increased the OPCache size.

PR #430, opened 2018-01-05
Show all products on manufacturer listing products

Zengraph:
If products has no image, it won't show on manufacturer product listing, changing SQL like in product class resolve the problem.


firstred:
Ahh that is a nice way of fixing it. Sorry for the long wait, this was hard to test.

firstred:
Approved and let's merge it. Thanks!

PR #435, opened 2018-01-10
Fix returning HTML error to AJAX request

mcdado:
When reordering images in a product with the multishop active, in some cases there might be exceptions, and this leads to HTML being returned to a AJAX request expecting JSON, with the following:
```
SyntaxError: JSON Parse error: Unrecognized token '<'
```
visible in browser console. With this, the response will always be JSON.


firstred:
Makes sense. Thank you!

PR #436, opened 2018-01-10
Fix Webservices Combinations

benzon:
Shop::addSqlRestriction - returned a faulty Webservice reply for combinations, changed to Shop::addSqlAssociation - and now returns expectet reply.


firstred:
That seems to do the trick. Thanks, mate!

benzon:
No problem 👍

PR #437, opened 2018-01-10
XML Output starter changed back to PS

benzon:
Helped out one on the forum today, he had issues with his service provider sendcloud, they use Webservices, and are looking for the prestashop opening tag in XML files - <prestashop xmlns:xlink="http://www.w3.org/1999/xlink">

But failed to import orders since the reply was <thirtybees xmlns:xlink="http://www.w3.org/1999/xlink">

Think it needs to be changed back to prestashop to keep compatibility alive when it comes to webservices.


firstred:
That's just what we needed. Going to test it in a second. Thanks!

firstred:
That's perfect. Merged!

PR #438, opened 2018-01-11
Fix reordering product images in multishop

mcdado:
This happens when some images are not enabled in the current shop.

To reproduce enable multishop and add a secondary shop, add images to a product, disable one of them in the current shop and then save the product. Now try to change the order of the images for that product. Since https://github.com/thirtybees/thirtybees/pull/435 you will visible see an error that says `Property Image->id_product is empty`. This PR fixes that error.

### DIFF
--- a/controllers/admin/AdminProductsController.php
+++ b/controllers/admin/AdminProductsController.php
@@ -1516,8 +1516,21 @@ class AdminProductsControllerCore extends AdminController
             $json = stripslashes($json);
             $images = json_decode($json, true);
             foreach ($images as $id => $position) {
-                $img = new Image((int) $id);
-                $img->position = (int) $position;
+                /*
+                 * If the the image is not associated with the currently selected shop, the fields that are also in the
+                 * image_shop table (like id_product and cover) cannot be loaded properly, so we have to load them
+                 * separately.
+                 */
+                $img = new Image((int)$id);
+                $def = $img::$definition;
+                $sql = 'SELECT * FROM `' . _DB_PREFIX_ . $def['table'] . '` WHERE `' . $def['primary'] . '` = ' . (int)$id;
+                $fields_from_table = Db::getInstance()->getRow($sql);
+                foreach ($def['fields'] as $key => $value) {
+                    if (!$value['lang']) {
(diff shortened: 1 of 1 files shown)


firstred:
Hey, thanks for your pull request.

I am not sure if this is the right solution, though. It calls `$img->update();` which is used to trigger ObjectModel update hooks. Just entering it in the database would skip this step and might break modules that depend on it.

Isn't the product ID known at the time we update the position? If so, wouldn't just setting the `id_product` property if it is missing solve it, too?

mcdado:
Good point. I guess that’s doable, I’ll look into it later tonight
-- 

--
*David Gasperoni*

firstred:
I pushed a few commits. Let me know if that solution, works, too. I think we can go with it, but not sure if `id_product` can always be grabbed with `Tools::getValue`

mcdado:
I looked into it, and that's not actually a good solution. First of all in my tests I don't have `id_product` as GET parameter. Second, it also doesn't load the `cover` field, so it needs to be loaded too.

I think the best way is to a read of all the fields in a single query (non join) and set them on the object. Maybe it could be done dynamically in case fields are added in the future. I'm looking into it right now.

mcdado:
So, first of all I rebased everything on top of `1.0.x`, and got rid of previous commits.

Now, I could have simply loaded the individual fields, but since you never know if there are going to be more fields, I wanted to pull them all in. That's also because if you don't pull in a field that is also present in `image_shop`, then when calling `->update()` that field would get set to null.

I also found out that there is a similar problem with the cover selection, so I applied the same logic there.

I also found out another problem:
1. you are in a shop context
2. disable that shop for an image
3. make that image the cover image
4. click to re-enable the shop for that image

and you'll run into an error. That's because `ajaxProcessUpdateProductImageShopAsso()` tries to insert the record in `image_shop`, but there's already a record with that `id_product`, `id_shop`, and `cover`. That goes beyond this PR, though.

Traumflug:
Thank you for the second commit.

I can confirm it solves the issue. Picked it to 1.0.x (with some whitespace edits).

For the other issue you spotted: well, another commit (or pull request) is welcome, of course!

PR #439, opened 2018-01-12
Removed submodule modules/statsbestmanufacturers

SLiCK-303:
remove modules/statsbestmanufacturers as it is built into statsmodule now


firstred:
I double checked and checked, but indeed it has to be removed. I thought it was still necessary. Thanks!

PR #448, opened 2018-01-16
Fix pagination on the customer group list

SLiCK-303:
When you try to navigate through differents pages of the customer group list, the pagination will redirect you to the groups list instead of the specified page.


SLiCK-303:
How to test: BO > Customers > Groups > choose a group, try to navigate the pagination and check if you will redirect to the correct page.

firstred:
Yes, it works and looks like a great solution. Thanks!

PR #449, opened 2018-01-16
Fix redirection when filtering product feature values

SLiCK-303:
When you try to search a feature value in the feature values list page , you will be redirected to the features list and the results won't be displayed.


SLiCK-303:
To reproduce: BO > Catalog > Product Features > View a feature and try to search by value, check if the search result is displayed correctly and not redirect to the features list.

firstred:
Yes, this solution is spot on. Thanks!

PR #450, opened 2018-01-16
Fix accesing array as an object

TobyMudito:
Bug trying to access array $products as an object.


firstred:
Thanks, going to check it right away. I can imagine there are cases where it's an object, but I am not sure.

firstred:
Yup, this fixes it. Thanks!

`Supplier::getProductsLite` always returns an array.

PR #451, opened 2018-01-17
Support favicon templates

firstred:
Adds a favicon template editor, which reduces the source images to just one which gets automatically resized according to the tags chosen:
faviconconfig


alwayspaws:
Very nice. Thank you!

PR #452, opened 2018-01-17
Fix for keywords disapear from search index

SLiCK-303:
Delete a product, and you delete words that is shared in search_word table, make other product disappear from search until next full reindex.

### DIFF
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -4463,12 +4463,18 @@ class ProductCore extends ObjectModel
     public function deleteSearchIndexes()
     {
         return (
-        Db::getInstance()->execute(
-            'DELETE `'._DB_PREFIX_.'search_index`, `'._DB_PREFIX_.'search_word`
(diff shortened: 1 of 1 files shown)


SLiCK-303:
From the administration panel, Catalog - > Products

This is the issue:
When deleting one product, all products with similar names, have been deleted from the search index, even if, going to Prefences -> Search, all products seem to be correctly indexed, but on the front end if I search for a product with a similar names of that deleted product, no records has been found.

Other products with different names than deleted one has not affected by this issue.

Re-building the entire search index fix the issue and all product are available for search again.

Example
Procuct 1 Name : Test product 1
Procuct 2 Name : Test product 2

Going on the frontend and searchig for "test" key, 2 records have been found

I delete Test Product 1

Going on the frontend and searchig for "test" key, no records have been found

Rebuilding the index and searching again for "test" , it display the product "test Product 2"

firstred:
Indeed, it is caused by that join that should not be there. Thanks for this pull request, it works. Great!

PR #458, opened 2018-01-21
Show tax product info

TobyMudito:
When adding via ajaxProcess a new product, this modification will show the product tax.


firstred:
That makes sense. Thank you!

PR #459, opened 2018-01-21
Add tax info when adding a new product

TobyMudito:
Currently, when you add a new product to a supply order, the tax value field in the template is always 0. With this help, helper template has the necessary info to display the associated tax.


firstred:
Approved! Thank you.

PR #467, opened 2018-02-26
Fix hreflang tag

daresh:
The hreflang tag should contain a language code, not iso code.
Iso code is incorrect, cause for example it would give a "gb" language which doesn't exist in the ISO spec. On the other hand using the language code would give us a "en-gb" value.


firstred:
Correct! Thank you!

PR #475, opened 2018-03-30
Switch prestashop links for tb ones in installer

gonssal:
I also sneaked in some catalan translations.


firstred:
Thanks a lot!

PR #510, opened 2018-05-10
php 7.2 support.

yaniv14:
in PHP 7.2
````count(): Parameter must be an array or an object that implements Countable````

uname param always return a string:
````https://github.com/thirtybees/thirtybees/blob/1.0.x/controllers/admin/AdminInformationController.php#L115````


Traumflug:
Thanks for the patch, picked it with https://github.com/thirtybees/thirtybees/commit/7dc1e9bfac0e66947b0967b1b8e334b2cff87f3a

... and now I have to fix branch *1.0.x*, because it no longer passes the testsuite due to my recent commits. One minute of sloppiness ...

Traumflug:
Everything passing now. Travis failure was apparently an infrastructure hiccup.

PR #520, opened 2018-05-26
Fix bad display manufacturers when multishop

okom3pom:


Traumflug:
Thanks for bringing up the issue.

The provided commit smells a bit like hiding but not fixing an underlying issue, which is why I investigated the area a bit.

Steps to reproduce:
- Enable multishop.
- Add a second shop.
- Go to back office -> Price Rules -> Catalog Price Rules.
- Edit a rule.
- Open the *Add a new condition group* panel, button at the bottom.
In the *Manufacturers* dropdown, the single manufacturer available is listed twice.

Investigation shows, this dropdown lists all manufacturers of all shops combined. Which means, if a manufacturer is associated to two shops, it's listed twice. Adding a *GROUP BY* clause, what the commit does, removes duplicates from the list, a side effect of this clause.

Price rules aren't associated to shops, so the IMHO right behavior would be to list all manufacturers, independent on which shop they're associated to. Will look up similar situations to see how code handles things there.

Traumflug:
After looking into the issue a bit more, I came up with this (branch *manufacturer*): https://github.com/thirtybees/thirtybees/commit/afe545371fa4c0375848b6a3cf6bd2f3ca6213e4

What do you think? If there are no complaints I'll forward this to *1.0.x* in a couple of days.

okom3pom:
It seems good and better

Traumflug:
Excellent. Moved this to branch *1.0.x* to get it into the next release.

Thank you very much!

PR #529, opened 2018-06-08
Import categories with delete option doesn't delete tables category_lang and category_shop

pelechano:
Import categories with delete option doesn't delete tables category_lang and category_shop


Traumflug:
Thank you very much for the contribution!

Picked the commit onto *1.0.x* to make sure it gets into the next release.

PR #544, opened 2018-06-26
Ensure RewriteCond for shop domain is always present

lucasRolff:
## Description 
When "Friendly URLs" are enabled together with Media Servers, there's a race condition where it will result in 404 for images either on the Media Server URL or the actual shop URL. 
The issue is caused by wrong rewrite rules and the lack of a rewrite rule matching the shop domain. The bug isn't directly apparent on Apache 2.4.x servers because Apache silently "fixes" the rules (RewriteCond ending with [OR] without a following RewriteCond). The fix is to include the shop domain.

## How to test

- Setup a shop on a Litespeed (apache compatible, except it doesn't contain a mod_rewrite bug)
- Enable a media server (such as `cdn.domain.com`)
- access an image on the shop URL such as `http://domain.com/thirtybees/3-home_default/soap.jpg` - this should work fine with my fixes as well

## Why this change should be implemented

Thirtybees should be compatible with Apache and Litespeed (a lot of shared hosts use this), the fact that current implementation has worked in Apache so far is simply due to a bug in the mod_rewrite engine by ignoring `RewriteCond` rules ending with `[OR]` when there has been no proceeding `RewriteCond`.

It also makes no sense to not include the shop URL in the RewriteCond, you'd want your images to be accessible on your media servers and on your shop URL at all times - the original code in reality shouldn't allow that.

The change is backwards-compatible and is a non-breaking change.

The benefit is that it will actually allow users of Litespeed to utilize friendly URL's and media servers at same time if these media servers happen to be a CDN it's where most people will see breakage.

The same bug has been submitted to Prestashop 1.6 and 1.7 - and it's in the process of getting merged into a bugfix release.

Since we have thirtybees users, we'd like the fix to exist in Thirtybees as well, thus I'm creating this PR to you guys.

If you have any questions or concerns, I'm always happy to answer them 👍

### DIFF
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -3288,21 +3288,15 @@ class ToolsCore
                 if ($uri['virtual']) {
                     if (!$rewrite_settings) {
                         fwrite($write_fd, $media_domains);
-                        if (Shop::isFeatureActive()) {
-                            fwrite($write_fd, $domain_rewrite_cond);
-                        }
+                        fwrite($write_fd, $domain_rewrite_cond);
                         fwrite($write_fd, 'RewriteRule ^'.trim($uri['virtual'], '/').'/?$ '.$uri['physical'].$uri['virtual']."index.php [L,R]\n");
                     } else {
                         fwrite($write_fd, $media_domains);
-                        if (Shop::isFeatureActive()) {
-                            fwrite($write_fd, $domain_rewrite_cond);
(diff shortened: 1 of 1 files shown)


Traumflug:
Looks good! Thanks you very much for the contribution. Picked the commit onto *1.0.x* to get it into the next release.

Issue #587, opened 2018-09-01
B2B siret/ape validation error


palo:
Siret and ape validation is incorrect for portuguese and probably many other locations.
Portuguese format example bellow. 

Steps to reproduce:

1. Enable b2b mode;
2. Add a customer from backoffice;
3. APE "PT999999999"
4. SIRET "70220"
5. Save

Plase fix Class/Validate or make validation optional.

Thank you.

Traumflug:
Thanks for the report.

Uhm. What is "Siret"? Or "ape"?

doclucas:
![](https://lh3.googleusercontent.com/-0iZ5SdbQbQM/W4qON8K_HwI/AAAAAAAAKvs/Zb5fnmWiWBMBKfdAIAFJJ5HwsH1Zc_C_QCHMYCw/s0/chrome_2018-09-01_20-03-50.png)
The SIRET code (French: Système d’identification du répertoire des établissements), or SIRET number, is an INSEE code which allows the geographic identification of any French establishment or business.
APE - http://www.startbusinessinfrance.com/blog/post/tip-what-is-your-ape-code

doclucas:
Related discussion:
https://www.prestashop.com/forums/topic/573067-vat-siret-ape-not-valid/
It seems adding a button that makes validation of these fields optional is indeed a good solution, or add modify the validation options to be compatible with other countries using these codes with different string formats.

doclucas:
Not the most elegant solution, but suggested fix:
https://github.com/thirtybees/thirtybees/pull/588

palo:
Still a bug in Thirtybees 1.1.0.
We really need a permanent fix for this.
Having a VAT field usable in any country is base requirement for an e-commerce platform.

getdatakick:
siret/ape are not for VAT numbers

palo:
Sure, but... Having an universal "vat number" tied to the address... 
While "siret/ape" (with validation for few countries) tied to the customer?

Also "siret/ape" shows right next to the company field that's ideal for b2b...
Maybe you'r overthinking this.
Its way better to make siret/ape validation optional. Just let the translation handle siret and ape to local equivalent name.

palo:
Closing this as a bug because:
- This is more a feature request than a bug;
- Siret/ape fields can be removed from theme template;
- Validation can be overwritten if required for a b2b project; 

Still think it should be optional, because siret doesn't apply to many countries (only France?).

[...]

PR #588, opened 2018-09-01
Suggestion solution for failing SIRET and APE codes out of France

Pharaoh2k:
Suggested solution for: https://github.com/thirtybees/thirtybees/issues/587
User needs to add a dash at the beginning of the strings to skip validation or simply enter empty string.

### DIFF
--- a/classes/Validate.php
+++ b/classes/Validate.php
@@ -1540,6 +1540,10 @@ class ValidateCore
      */
     public static function isSiret($siret)
     {
+        if (empty($siret) || (isset($siret) && $siret[0] == '-')) 
+        { 
+            return true; 
+        }
         if (mb_strlen($siret) != 14) {
             return false;
         }
@@ -1567,6 +1571,9 @@ class ValidateCore
      */
     public static function isApe($ape)
     {
+        if (empty($ape) || (isset($ape) && $ape[0] == '-')) { 
+            return true; 
+        }
         return (bool) preg_match('/^[0-9]{3,4}[a-zA-Z]{1}$/s', $ape);
     }
 
--- a/controllers/admin/AdminCustomersController.php
+++ b/controllers/admin/AdminCustomersController.php
@@ -637,11 +637,14 @@ class AdminCustomersControllerCore extends AdminController
                 'type'  => 'text',
                 'label' => $this->l('SIRET'),
                 'name'  => 'siret',
+				'hint'  => $this->l('To skip validation, add a dash at the beginning of the SIRET code'),
             ];
+
             $this->fields_form['input'][] = [
                 'type'  => 'text',
                 'label' => $this->l('APE'),
                 'name'  => 'ape',
+				'hint'  => $this->l('To skip validation, add a dash at the beginning of the APE code'),
             ];
             $this->fields_form['input'][] = [
                 'type'  => 'text',


getdatakick:
I don't think this is a correct fix.

First of all, it allows for entering empty siret / ape codes. But I believe (in France at least) this information is a must have for B2B customers.

The main problem is that these fields exists in core at all. They were designed specifically for french companies, and that's just wrong. Other countries should NOT use them, even though they have some similar company identification codes. They are simply not [SIRET](https://en.wikipedia.org/wiki/SIRET_code) codes.

doclucas:
The reason I added an option to enter empty string is exactly so that those fields can be skipped in case those codes are not needed and thus were not entered.
Of course it's possible to add a condition for that empty option to work only for business that are out of France and show notice otherwise.

As for the second point, having custom fields for specific countries/segments in the core itself isn't really  wrong per se IMHO.
I guess a better implementation of such country/segment-specific would have been to get them to be visible and required when actually needed (ie. for businesses in France) and not for everybody. Note that they appear also in the FO.

getdatakick:
> The reason I added an option to enter empty string is exactly so that those fields can be skipped in case those codes are not needed and thus were not entered.

I've just tested it, and it's possible to leave these fields empty even now. The validation is performed only when these fields contain some value. 

That means that the proposed fix has no affect at all. In only allows for `-` to be used as siret/ape value as well, which is not much helpful. It does not solve the original issue - Portugal version of *siret* will not pass the validation.

To fix this, we would either need to drop the validation completely, or implement country-specific validation. The later is quite a large task, because validator function in `Validate` class are not context-sensitive, and we need context here.

doclucas:
>That means that the proposed fix has no affect at all. In only allows for - to be used as siret/ape value as well, which is not much helpful. It does not solve the original issue - Portugal version of siret will not pass the validation.

So the `empty` is redundant, can be removed.
The `-` is actually helpful, as it make sure the string skips validation completely, so that Portugal version can be entered without any issues, with that preceding dash. Of course it's not the most optimal way to do it, I wrote that before, having a proper validation algorithm for each country is much better, but this is much better (temporary fix at least) than it is now when they have no way to enter non-french SIRET/APE codes at all.

getdatakick:
It just seems very strange to me. Why should there be a validation, if the validation can be bypassed? Isn't it better to skip the validation altogether, since it no longer serve its purpose?

Traumflug:
A bypassable validation protects mostly against typos and such stuff.

That said, what's the problem with making the validation localized? Just a bit more work?

doclucas:
Obviously the validation here is a not a security feature, it's to check and prevent typos, which can be made optional.

getdatakick:
> Obviously the validation here is not a security feature, it's to check and prevent typos, which can be made optional.

Well, I don't know. Maybe there is some french merchant that actually depends on these fields to contain well-formatted and validated values. Maybe they copy and paste them to some bullshit tax return form, or whatnot. 

If we add an option to skip the validation, then we are breaking it. It's probably quite far-stretched, but it's possible.

> That said, what's the problem with making the validation localized? Just a bit more work?

yes, it's just a bit more work. We would need to somehow define context-aware validators, and fix the code that actually [calls](https://github.com/thirtybees/thirtybees/blob/13ced3c168470ef518bd87facf02e94ebe2008ea/classes/ObjectModel.php#L1179-L1204) validator functions to provide the context

doclucas:
>Well, I don't know. Maybe there is some french merchant that actually depends on these fields to contain well-formatted and validated values. Maybe they copy and paste them to some bullshit tax return form, or whatnot.
>If we add an option to skip the validation, then we are breaking it. It's probably quite far-stretched, but it's possible.

That's why it's called "an option", if the validation is important for that merchant, he wouldn't use the option to skip it. If he does, nothing we can or should do with self inflicted issues.

getdatakick:
> That's why it's called "an option", if the validation is important for that merchant, he wouldn't use the option to skip it. If he does, nothing we can or should do with self inflicted issues.

But it's the customer who enters these data.

doclucas:
I am talking specifically about the BO side, for which this PR is meant for.
I realize the "dash-to-skip" in the function affects the FO as well, but customers won't know it and I already said it's not perfect. Adding an option for the admin to decide if he wants to disable validation (in case he doesn't rely on it for any bureaucratic reason) is preferred of course, or whatever you guys think is better. Feel free to do what you think is right.

Traumflug:
Thinking about this a bit and looking at the *vatnumber* module, these APE and SIRET numbers would be a good candidate for going into a module. Like it's done for vat numbers.

One can set the flavor of this number on the module configuration page, then. This also allows allows to add some online verification tool.

Quite some work, though. As I'm not exactly in love with adding such magic characters (`-`), is there some way to get number validation matching French as well as Portugese ones? Like `return (strlen($string) === 12 || strlen($string) === 14)` ?

ecastroba:
In my case, I set up a regular expression for the two fields to work in Colombia (SIRET = tax identification number, APE = chamber of commerce registration).

At classes/Validate.php

`   public static function isSiret($siret)
    {
    	return (bool) preg_match('/^[0-9]{1,15}-[0-9]{1}$/s', $siret);
    }

    public static function isApe($ape)
    {
        return (bool) preg_match('/^[0-9]{1,10}$/s', $ape);
    }`

Maybe if both parameters could be moved to some configuration page or module, each one could write their own regular expression.

PR #597, opened 2018-09-13
Rindjael with PHP 7 - SSL encrypt

Captain-FLAM:
For explanations, see there :

https://forum.thirtybees.com/topic/1937/rindjael-with-php-7-ssl-encrypt

### DIFF
--- a/classes/Rijndael.php
+++ b/classes/Rijndael.php
@@ -72,13 +72,11 @@ class RijndaelCore
         }
 
         if (function_exists('openssl_encrypt')) {
-            $ivsize = mcrypt_get_iv_size(MCRYPT_RIJNDAEL_128, MCRYPT_MODE_CBC);
+            $ivsize = openssl_cipher_iv_length('aes-256-cbc');
             try {
                 $iv = random_bytes($ivsize);
             } catch (Exception $e) {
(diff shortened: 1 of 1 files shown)


getdatakick:
Thanks, this is definitely a good fix, even though there's a plan to get rid of Rindjael class in the future.

I've modified your code a bit - because we know that openssl extension is loaded, we can directly use `openssl_random_pseudo_bytes`, and not depend on `random_bytes`, which exists on php7 only.

Here's the commit: 5e416764cd11708ffebb7109f8a86edfe66c2973

Captain-FLAM:
ok, good, so I delete my branch ...
:+1: 

> Thanks, this is definitely a good fix, even though there's a plan to get rid of Rindjael class in the future.
I'm a little bit afraid about changing crypto on existing and running TB ... I hope everything will go well, we'll see ...

getdatakick:
> I'm a little bit afraid about changing crypto on existing and running TB ... I hope everything will go well, we'll see ...

Thank you again for proposing the fix. 

If you are using Rindjael encryption and your server has openssl installed, you can continue to use this encryption, nothing will change. There shouldn't be any issues.

PR #599, opened 2018-09-18
Avoid unnecessary DB query on users' password check.

gonssal:
When a user login attempt fails, avoid an unnecessary database query.

Also check that the user-sent password is a valid md5 hash before trying to hash and check it.


Traumflug:
I love seeing unnecessary code to get removed!

Thank you very much, picked to *1.0.x*.

PR #600, opened 2018-09-18
Double definition

Captain-FLAM:
Not comment ... just look at it


Traumflug:
Nice catch!

However, the file you changed is a vendor file, a file just copied from another project. We usually don't edit these.

Is there a pressing issue for this change, like some malfunction? If not, I'd like to postpone such changes to next year, when we can begin to remove old cruft. thirty bees comes with no less than 604 jQuery files and probably all of them are several years old. A situation which undoubtly needs some cleaning.

To not forget this, I've files #602.

Captain-FLAM:
:+1:

PR #627, opened 2018-10-15
CO : Error when post_max_size is 0

idnovate:
| Questions     | Answers
| ------------- | -------------------------------------------------------
| Description?  | If you configure `post_max_size = 0` in php.ini and try to edit a product, there's a JS error in files `admin/themes/default/template/controllers/products/helpers/uploader/attachment_ajax.tpl` and `admin/themes/default/template/controllers/products/helpers/uploader/ajax.tpl` because variable `$post_max_size` is empty.
| Type?         | bug fix
| Category?     | CO 
| BC breaks?    | no
| Deprecations? | no
| How to test?  | Set `post_max_size = 0` in php.ini. Try to edit a product.

### DIFF
--- a/classes/Uploader.php
+++ b/classes/Uploader.php
@@ -284,6 +284,10 @@ class UploaderCore
     public function getPostMaxSizeBytes()
     {
         $postMaxSize = ini_get('post_max_size');
+        if (!$post_max_size) {
+            return static::DEFAULT_MAX_SIZE;
+        }
+        
         $bytes = (int) trim($postMaxSize);
         $last = strtolower($postMaxSize[strlen($postMaxSize) - 1]);
 
--- a/classes/Uploader.php
+++ b/classes/Uploader.php
@@ -284,7 +284,7 @@ class UploaderCore
     public function getPostMaxSizeBytes()
     {
         $postMaxSize = ini_get('post_max_size');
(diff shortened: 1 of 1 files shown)


Traumflug:
You didn't test your patch, did you?

Variable names mismatch. `$postMaxSize` isn't the same as `$post_max_size`.

idnovate:
@Traumflug I tested the patch at PrestaShop and then copied it to ThirtyBees, and variable names mismatch (https://github.com/PrestaShop/PrestaShop/blob/1.6.1.x/classes/Uploader.php#L117)

Traumflug:
Fine. Tested the issue and can confirm it. Nice catch, thank you!

Instead of adding code, I found a solution by removing code, see https://github.com/thirtybees/thirtybees/commit/6bc46fe1ba998efbf7b5ee2b68cd583632f81370

Does this work for you as well?

Returning `DEFAULT_MAX_SIZE` would change the meaning of this method. If this is actually wanted, e.g. for actual upload limits, it'd need more adjustments. Like handling the case where `DEFAULT_MAX_SIZE` is 4MiB, while PHP's `post_max_size` is 8 MiB.

Traumflug:
Apparently no objections. Moved this stuff to *1.0.x*. Thank you very much for the contribution!

PR #628, opened 2018-10-15
CO : Fixed error with redirections

idnovate:
https://www.prestashop.com/forums/topic/910826-redirection-https-vers-http-vers-https-si-www/

### DIFF
--- a/classes/shop/Shop.php
+++ b/classes/shop/Shop.php
@@ -370,7 +370,7 @@ class ShopCore extends ObjectModel
                         $redirectHeader = ($redirectType == 1 ? 'Found' : 'Moved Permanently');
                         header('HTTP/1.0 '.$redirectCode.' '.$redirectHeader);
                         header('Cache-Control: no-cache');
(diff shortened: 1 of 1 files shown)


Traumflug:
Origin by @jeckyl: https://github.com/PrestaShop/PrestaShop/pull/11004/commits/b213fe7844bd621c38dcdc2ad1dbeabba373ba52

Dh42:
I agree, that is so much faster than having it travel through php.

Traumflug:
I'd be happier if I could reproduce something not working ideally. Without the patch everything works just fine

`http://localhost/thirtybees`
redirects 301 to
`https://localhost/thirtybees/`
redirects 302 to
`https://localhost/thirtybees/en/`

Traumflug:
@Zengraph found a simpler solution for this, which entered *1.0.x*: f1877ee1baeabf973b825ef59e47b482b69fb6c4

Thank you very much for the heads up, anyways.

Issue #629, opened 2018-10-15
Backoffice data filter


mauroagr:
When i filter a data in orders or customers list, and before i have make others filters, the 'chrome suggestion', stay over the calendar like the image below.
![image](https://user-images.githubusercontent.com/5348884/46976556-26da8500-d0a0-11e8-88ad-3871803ef56c.png)

Dh42:
This is some junk the new chrome added. I am unsure if we can do anything about it currently.

mauroagr:
I find some solution, will test, using autocomplete="off" in the field.

PR #630, opened 2018-10-15
Fix autocomplete in datepicker fields - #629

mauroagr:
In google chrome in date fields they suggest old filtered dates, and put over the datepicker. This solve the problem


Traumflug:
Picked to *1.0.x*. Thank you very much for the contribution!

PR #636, opened 2018-10-25
Fix TB_CUSTOMCODE_METAS was not actually saved

TrueXakeP:
Different sanitization between `pSQL()` and `Configuration::set()` make the situation where HTML (together with meta tags in _METAS) were removed prior to the value was updated in the database but kept on the page/editor as if it was saved. And so it is difficult to notice the loss on time.
So I decide also to move tags stripping prior to booth `pSQL()` and `Configuration::set()` while allowing meta tags.
And I see a similar thing in the `Configuration::updateValue()`.

Should other tags be allowed in _METAS in case of the info note on the page says about any HTML in `<head>`?

HTMLPurifier's love of breaking complex css (and possibly another) with `&gt;`/ `&lt;` along with the mentioned and independent tag cleaning (in `pSQL`) makes this warning a little inaccurate.
https://github.com/thirtybees/thirtybees/blob/076aaa5d56d4eb6d76b10aae3eda919f90973db6/controllers/admin/AdminCustomCodeController.php#L28

// Sorry for my English.

### DIFF
--- a/controllers/admin/AdminCustomCodeController.php
+++ b/controllers/admin/AdminCustomCodeController.php
@@ -113,8 +113,7 @@ class AdminCustomCodeControllerCore extends AdminController
         }
         $called = true;
 
-        $safeMetas = strip_tags(Tools::getValue(Configuration::CUSTOMCODE_METAS), '<meta>');
-        $this->updateOptionUnescaped(Configuration::CUSTOMCODE_METAS, $safeMetas, true);
+        $this->updateOptionUnescaped(Configuration::CUSTOMCODE_METAS, Tools::getValue(Configuration::CUSTOMCODE_METAS), true);
     }
 
     /**
@@ -150,7 +149,7 @@ class AdminCustomCodeControllerCore extends AdminController
         }
         $called = true;
 
-        $this->updateOptionUnescaped(Configuration::CUSTOMCODE_ORDERCONF_JS, Tools::getValue(Configuration::CUSTOMCODE_ORDERCONF_JS));
+        $this->updateOptionUnescaped(Configuration::CUSTOMCODE_ORDERCONF_JS, Tools::getValue(Configuration::CUSTOMCODE_ORDERCONF_JS), true);
     }
 
     /**
--- a/controllers/admin/AdminCustomCodeController.php
+++ b/controllers/admin/AdminCustomCodeController.php
@@ -167,7 +167,7 @@ class AdminCustomCodeControllerCore extends AdminController
         }
         $called = true;
 
-        $this->updateOptionUnescaped(Configuration::CUSTOMCODE_JS, Tools::getValue(Configuration::CUSTOMCODE_JS));
+        $this->updateOptionUnescaped(Configuration::CUSTOMCODE_JS, Tools::getValue(Configuration::CUSTOMCODE_JS), true);
     }
 
     /**
--- a/controllers/admin/AdminCustomCodeController.php
+++ b/controllers/admin/AdminCustomCodeController.php
@@ -173,11 +173,12 @@ class AdminCustomCodeControllerCore extends AdminController
     /**
      * @param string $key
      * @param string $value
+     * @param bool   $htmlOK
      *
      * @throws PrestaShopDatabaseException
      * @throws PrestaShopException
      */
-    protected function updateOptionUnescaped($key, $value)
+    protected function updateOptionUnescaped($key, $value, $htmlOK = false)
     {
         if (Db::getInstance(_PS_USE_SQL_SLAVE_)->getValue(
             (new DbQuery())
@@ -188,7 +189,7 @@ class AdminCustomCodeControllerCore extends AdminController
             Db::getInstance()->update(
                 bqSQL(Configuration::$definition['table']),
                 [
-                    'value'    => pSQL($value),
+                    'value'    => pSQL($value, $htmlOK),
                     'date_upd' => ['type' => 'sql', 'value' => 'NOW()'],
                 ],
                 '`id_shop` IS NULL AND `id_shop_group` IS NULL AND `name` = \''.pSQL($key).'\'',
@@ -200,7 +201,7 @@ class AdminCustomCodeControllerCore extends AdminController
                 bqSQL(Configuration::$definition['table']),
                 [
                     'name'          => pSQL($key),
-                    'value'         => pSQL($value),
+                    'value'         => pSQL($value, $htmlOK),
                     'id_shop'       => null,
                     'id_shop_group' => null,
                     'date_add'      => ['type' => 'sql', 'value' => 'NOW()'],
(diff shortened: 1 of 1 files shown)


Traumflug:
First, thanks for the contribution!

Could you give some samples of custom code which didn't work before, but works with your patches? This should allow to better understand what the issue is.

TrueXakeP:
Thanks for attention. And for your work.

I just put simple meta tag to the custom metas as I was trying to set up Google's site ownership verification. Just this one: `<meta name="google-site-verification" content="miaGFagvf2vKHKCzbF53uLKNE0mT-YlDVD6T0JVs7rs" />`.
But after applying changes it was not appeared on the front office, although he remained in the editor.
My investigation led me to the `$htmlOK=false` as `pSQL()`'s parameter and to the `strip_tags()` behind of it.
https://github.com/thirtybees/thirtybees/blob/076aaa5d56d4eb6d76b10aae3eda919f90973db6/config/alias.php#L69-L71
https://github.com/thirtybees/thirtybees/blob/076aaa5d56d4eb6d76b10aae3eda919f90973db6/classes/db/Db.php#L869-L870


This note says that I can use "extra HTML" to put it in the a `<head>` and I assume that any valid there:
https://github.com/thirtybees/thirtybees/blob/076aaa5d56d4eb6d76b10aae3eda919f90973db6/controllers/admin/AdminCustomCodeController.php#L85
However this one tells only about `<meta>` tags:
https://github.com/thirtybees/thirtybees/blob/076aaa5d56d4eb6d76b10aae3eda919f90973db6/controllers/admin/AdminCustomCodeController.php#L34

Also I assume that without allowing tags in custom JS something like this may be broken: `$("").attr('src', var);`.
Maybe for custom CSS this all will be irrelevant.

TrueXakeP:
In short this code as custom "meta tags" will wrongly saved but will be kept in the editor without my patch:
```html
<meta name="google-site-verification" content="miaGFagvf2vKHKCzbF53uLKNE0mT-YlDVD6T0JVs7rs" />
<link rel="apple-touch-icon" href="/img/app_icon.png">
<script type="text/javascript" charset="utf-8">
    console.log('test');
</script>
```
And with my patch it will be turned to this and saved and shows in the editor the same:
```html
<meta name="google-site-verification" content="miaGFagvf2vKHKCzbF53uLKNE0mT-YlDVD6T0JVs7rs" />


    console.log('test');

```

As I assumed custom JS from this one `$('').attr('src', '/favicon.ico');` turned into this one `$('').attr('src', '/favicon.ico');`. But it is reflected in the editor immediately.

TrueXakeP:
I just found that I could place my favicons as custom metas in a `<head>` instead of putting them into the 'head.tpl'.
And that something like `<script src='https://....'>` in the head can be the valid case too.

In that case should be description of custom metas updated?

Traumflug:
Hmm. Previous Travis CI job pending since two hours. Restart doesn't help. https://travis-ci.org/thirtybees/thirtybees/builds/447352015?utm_source=github_status&utm_medium=notification

This one pending as well: https://travis-ci.org/thirtybees/thirtybees/builds/447393353?utm_source=github_status&utm_medium=notification Looks like this has to wait for a while.

Anyways, reviewed commits are on branch *customcode*, so they'll eventually end up on *1.0.x*. Let me say thank you already, this is a nice catch and a well written fix.

TrueXakeP:
The point of third commit (the https://github.com/thirtybees/thirtybees/pull/636/commits/f08f9fc0b08a26a624016de6e738b074a7fa1158 I guess) is to make stripping of tags visible to user after it press the save button.

TrueXakeP:
Oh I see now... That commit affect to the all customcode fields instead of first two.

Traumflug:
`updateOptionUnescaped()` is the wrong place for preparing data for display. Preparing data for display happens in `__construct()`.

I see that code fills property 'value' for two of the fields, but not for the other two fields. About what 'value' means, see http://jump-ing.com/traumflugs/helper_field_text (I hope it's right there). Still I'll have to look up whether displayed data gets actually fetched freshly from the database or if the HTML request data gets taken. Happens somewhere deep in the infrastructure.

Traumflug:
Values get set here: https://github.com/thirtybees/thirtybees/blob/1.0.x/classes/helper/HelperOptions.php#L346 Values coming with the request get favored over values in the database (which I consider to be an unfortunate strategy).

Accordingly, values should get set for all four fields. I'll prepare a commit in a minute.

Traumflug:
Here we go: https://github.com/thirtybees/thirtybees/commit/1875007550c230102c77b0d41a7c42fe6dd0abe8

TrueXakeP:
As I was thinking, the data for display is taken from within the https://github.com/thirtybees/thirtybees/blob/076aaa5d56d4eb6d76b10aae3eda919f90973db6/controllers/admin/AdminCustomCodeController.php#L245
which then process it differently then in `pSQL()`.
Yes I see my patch is not perfect in that side.

> 
> 
> Here we go: [1875007](https://github.com/thirtybees/thirtybees/commit/1875007550c230102c77b0d41a7c42fe6dd0abe8)

Hmm. It doesn't seem to work for me for the `$('').attr('src', '/favicon.ico');` as the custom JS. `` sanitized in DB but not in the editor.

Traumflug:
> I was thinking that this is by design that the data for display is not going trough the db

Sooner or later these values have to get loaded from the DB. When visiting this page from elsewhere, there is only the database. Re-using values from the HTTP request probably wants to save the time needed for the DB query.

As we see, this can lead to inconsistent behavior. This example code isn't sanitized when clicking just 'Save', as you noted. But when going to the back office menu, selecting *Preferences* -> *Custom Code* again, it is.

Digging into this issue a bit, the not sanitized value apparently gets cached by class *Configuration*, then sanitized, then stored into the DB. This makes the cache of this class dirty (different value than in the DB). Urps. Interesting comments in the code: https://github.com/thirtybees/thirtybees/blob/1.0.x/classes/Configuration.php#L610-L613

Probably not easy to get out of this mess without risking to break themes and modules.

Traumflug:
Digging deeper and deeper into the issue, eventually quite a number of commits were made. Just pushed them to branch *customcode*.

It's not complete, yet. But it looks much like all this configuration saving code in *CustomCodeController* can go away once class *Configuration* is fixed. More details in the commit messages.

It's midnight here, more to come tomorrow :-)

Traumflug:
Finally I managed to fix class *Configuration* to the point of escaping HTML values properly, see [branch customcode](https://github.com/thirtybees/thirtybees/commits/customcode).

This allowed to remove all this custom code in *CustomCodeController*, things still work nicely here. 162 lines of code less **\o/**. Does this work for you as well?

TrueXakeP:
Ok, I'll check this in a few days.

Traumflug <notifications@github.com> 11 ноября 2018 г. 7:46:01 PM написал:

Traumflug:
As I'm fairly confident with my work, I've moved it to branch *1.0.x* already. Please test it there.

Traumflug:
Assuming it's solved. If not, please don't hesitate to open another issue.

PR #640, opened 2018-10-27
Redirection to forced http after redirection domain (https)

Zengraph:
When the user requests the shop by one of the non canonical domains (domain.tld vs www.domain.tld),
PrestaShop causes a redirection to the http: //www.domain.tld.
This leads to a cascade of redirects. https: //domain.tld => http: //www.domain.tld => https: //www.domain.tld

source : https://aide.prestashop.click/topic/1068/1-5-0-17-redirection-vers-http-forc%C3%A9e-suite-%C3%A0-redirection-shop-domain-en-https


Traumflug:
Picked to *1.0.x*. Thank you very much!

PR #648, opened 2018-10-30
BO: Combination dropdown visibility

idnovate:
When you load a cart rule with a gift product and the gift product search found more than 1 product, all the combination dropdowns are visible.

| Questions     | Answers
| ------------- | -------------------------------------------------------
| Description?  | Only the product's combination dropdown is displayed
| How to test?  | Create a product with name and reference "AAA" with 2 combinations. Create a product with name and reference "AAAA" with 2 combinations. Create a cart rule with "AAA" as gift product. Save the rule. Load the rule and you will see 2 dropdowns, one with "AAA" combinations and another with "AAAA" combinations.

### DIFF
--- a/admin-dev/themes/default/template/controllers/cart_rules/form.js
+++ b/admin-dev/themes/default/template/controllers/cart_rules/form.js
@@ -501,6 +501,8 @@ $(document).ready(function() {
 			callback: function(text) { combinable_filter('#cart_rule_select_2', text, 'selected'); }
 		});
 	}
+	
+	displayProductAttributes();
 });
 
 


Traumflug:
@Dh42 and me tried to reproduce this issue in thirty bees, but we can't. Gift creation page looks just fine:

grafik

Are you sure it is an issue?

idnovate:
Hi!

It fails:

grafik

You have to:

- Create product with name and reference AAA. 2 combinations.
- Duplicate this product with name an reference AAAA.

Create a rule with AAA as gift. Save. Load this rule and then the issue appears.

Traumflug:
Aaah, one has to save. Now I see it as well. It's gone with your patch, great!

Picked it to *1.0.x*, see https://github.com/thirtybees/thirtybees/commit/22ea524d0feba6cddd5ddccf50f7591496f14daf, thank you very much for your contribution.

PR #773, opened 2018-11-10
Fixing two exploitable context-sensitive XSS vulnerabilities.

asteinhauser:
First and third video in https://github.com/asteinhauser/PrestaShop/files/2568663/prestashop.zip
Those vulnerabilities were in the PrestaShop codebase already when thirtybees forked from it.


Traumflug:
Thanks for the contribution.

- Installer fixes applied with 752499cf8d90fbf9212b976707d55941cc43a08a (independent commit, cherry-picking didn't work).
- Issue in `Helper->renderShopList()` isn't reproducible and no longer applies, because this method is deprecated since 1.0.0, no longer in use and got replaced by a method which doesn't assemble HTML directly.

If there's something left, please don't hesitate to re-open this pull request or file a regular issue.

PR #794, opened 2018-12-13
product line and ecotax visibility and correction

Zengraph:
Show ecotax on backoffice order detail, each line product show ecotax under base price and ecotax was added 2 times, now it's correct.


Traumflug:
Applied with a2767f304f71735e5fa6100abed3580628201fe3, thank you for the contribution. Used `displayPrice()` rather than a fixed formatting to 2 decimal places.

As always, if there's something left, don't hesitate to re-open this request or file a regular issue.

PR #814, opened 2019-01-15
Debug mode

getdatakick:
@Traumflug please review and give me your opinions on this change

Current implementation of debug mode depends on [display_errors](http://php.net/manual/en/errorfunc.configuration.php#ini.display-errors) settings. That means that all errors (including warnings and notices) are printed to the screen as part of the output. 

This behavior can break HTML markup significantly. But what's even worse, these error messages are even printed into ajax call response. If ajax call is supposed to return JSON data then such *enhanced* response can't be parsed.

Code in this branch is a proposed solution for this problem. It disable display_errors, and collect all errors and warnings instead (note that this was already implemented in back office). Collected error messages are then presented to user using [php-debugbar](https://github.com/maximebf/php-debugbar), both on backend and on frontend.

Using debugBar, we will be able to collect and display to user variety of interesting data in the future, such as sql queries, cache performance, etc... 

If you want to test the code, you will need to install new composer packages.

debugbar

### DIFF
--- a/classes/ErrorHandler.php
+++ b/classes/ErrorHandler.php
@@ -101,7 +101,8 @@ class ErrorHandlerCore
         // Set error handler
         set_error_handler([$this, 'errorHandler']);
 
-
+        // register shutdown handler to catch fatal errors
+        register_shutdown_function([$this, 'shutdown']);
 
         $this->initialized = true;
     }
@@ -164,8 +165,26 @@ class ErrorHandlerCore
         if (! $suppressed) {
             $this->logMessage($error);
         }
-
         return $suppressed || $this->preventDefaultErrorHandler;
+    }
+
+    /**
+     * Shutdown handler let us detect and react to fatal errors
+     */
+    public function shutdown()
+    {
+        $error = error_get_last();
+        if (static::isFatalError($error['type'])) {
+            $stack = [
+                1 => [
+                    'file' => $error['file'],
+                    'line' => $error['line'],
+                    'type' => 'Fatal error'
+                ]
+            ];
+            $exception = new PrestaShopException($error['message'], 0, null, $stack, $error['file'], $error['line']);
+            $exception->displayMessage();
+        }
     }
 
     /**
@@ -267,6 +286,20 @@ class ErrorHandlerCore
                 return 'Unknown error';
         }
     }
+
+    /**
+     * Returns true if errno is a fatal error
+     *
+     * @return boolean
+     *
+     * @since   1.0.9
+     * @version 1.0.9 Initial version
+     */
+    public static function isFatalError($errno)
+    {
+       return ($errno === E_USER_ERROR || $errno === E_ERROR);
+    }
+
 
     /**
      * Returns error PSR log level for given error level
--- a/classes/ErrorHandler.php
+++ b/classes/ErrorHandler.php
@@ -1,0 +1,79 @@
+<?php
+/**
+ * 2007-2016 PrestaShop
+ *
+ * thirty bees is an extension to the PrestaShop e-commerce software developed by PrestaShop SA
+ * Copyright (C) 2017-2018 thirty bees
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.txt.
+ * It is also available through the world-wide-web at this URL:
+ * http://opensource.org/licenses/osl-3.0.php
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@thirtybees.com so we can send you a copy immediately.
+ *
+ * DISCLAIMER
+ *
+ * Do not edit or add to this file if you wish to upgrade PrestaShop to newer
+ * versions in the future. If you wish to customize PrestaShop for your
+ * needs please refer to https://www.thirtybees.com for more information.
+ *
+ * @author    thirty bees <contact@thirtybees.com>
+ * @copyright 2017-2018 thirty bees
+ * @license   http://opensource.org/licenses/osl-3.0.php  Open Software License (OSL 3.0)
+ *  PrestaShop is an internationally registered trademark & property of PrestaShop SA
+ */
+
+/**
+ * Class ErrorHandlerCore
+ *
+ * @since 1.0.9
+ */
+class ErrorHandlerCore
+{
+
+    /**
+     * @var ErrorHandlerCore singleton instance;
+     */
+    protected static $instance;
+
+    /**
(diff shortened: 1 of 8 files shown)


Traumflug:
Doesn't this mean there should be all supported versions (5.6, 7.0, 7.1, 7.2, 7.3) listed? https://github.com/thirtybees/thirtybees/pull/814/commits/85dc0573cc6cce5c99e6a195845660c43d548933

I'd plan to drop PHP 5.5 support (and add 7.3 support) soon after finishing GitUpdater, so no need to work on this support.

getdatakick:
Composer libraries can be different for different php version. When you run `composer install`, it will ,by default, fetch library that matches your php cli version. For local development that's ok. But when we want to pack thirtybees to distribution zip file, we need to be sure that it contains library compatible with lowest php version supported. Which is, at the moment, 5.5.

For example look at package `paragonie/random_compat`. For php7 it will install a dummy library, because random_bytes polyfill is not required on this platform. But for php5 this polyfill is required. When we create thirtybees.zip file, this polyfill library must be there, otherwise php5.x users would be screwed. If you remember, this problem already happened once. The 85dc057 is here to prevent this in the future

getdatakick:
Alternatively, we could have different distribution zip files for different platforms, ie. `thirtybees_v1_0_9_php5.zip` and `thirtybees_v1_0_9_php7.zip`. 

That way, php7 users would benefit from optimized libraries for their platform.

Traumflug:
Glancing over all this: looks good!

Now looking a bit closer:
- [ ] Impact on release package size.
- [ ] Whether errors still appear without explicitly asking for them (else they'll get forgotten).
- [ ] How well it works with FireFox' debug console.
- [ ] How well it works with the currently existing footer bar.

And another question: what about current debug features, like `ddd()`? Can this get removed?

Traumflug:
> Composer libraries can be different for different php version.

Yes, that's what I assumed. But you list only PHP 5.5 there, not PHP 5.6 or PHP 7.x.

Dh42:
What about incrementing out of 1.0.x to 1.1.x and dropping 5.x support totally? I wonder if a lot of people would be affected.

getdatakick:
| Yes, that's what I assumed. But you list only PHP 5.5 there, not PHP 5.6 or PHP 7.x.

That's because the release zip file must be compatible with php5.5. If I listed other php versions there, then `composer` would choose what library to install according to php cli version. So, if we build release package on server with php7 installed, resulting zip file would contain php code **not** compatible with php5.5. 

When we put only the lowest supported version here, we don't give composer any wiggle room

Traumflug:
Answers:

Impact on release package size is 880 kB, quite a bit.

----
With or without debug mode, PHP errors still appear, like before:

grafik

grafik

----
Instead of a debug bar, I get errors in the JS console:

bildschirmfoto vom 2019-01-15 16-14-57
("Laden fehlgeschlagen" = loading failed)

These files failing to load exist and have read permissions. Well, except that they don't exist at `vendor/...`, but at `thirtybees/vendor/...`, there `thirtybees/` is the shop installation location.

----
DebugBox failing also breaks back office menu, which is not good. Loading DebugBox JS later/at last might solve this.

----
While I like the idea of directing errors out of the default location, a more lightweight approach might be a good idea. In some places, these Smarty errors appear in a red box like regular handling errors. Not sure where and why this happens, but it's certainly much less scary than the usual handling.

Another idea might be to simply forward PHP trouble to the JavaScript console, like shown [here](https://stackoverflow.com/a/20147885) or [here](https://stackoverflow.com/a/21586930) or [here](https://bueltge.de/einfaches-php-debugging-in-browser-console/): https://bueltge.de/einfaches-php-debugging-in-browser-console/. Doable with like 10 lines of code.

getdatakick:
> Impact on release package size is 880 kB, quite a bit.

Nobody will care that the zip file is a little bit bigger
 
> With or without debug mode, PHP errors still appear, like before:

This seems to be related to xdebug, I'll look into it. Without xdebug, error messages are displayed like this

errmsg

> Instead of a debug bar, I get errors in the JS console:

Good catch. I tested the on domain, you run tb from subdirectory. We need to pass relative url to [debugBar](https://github.com/maximebf/php-debugbar/blob/3169ed558deb4c8046f5421fa0240324f271ba1c/src/DebugBar/JavascriptRenderer.php#L212)

> DebugBox failing also breaks back office menu, which is not good. Loading DebugBox JS later/at last might solve this.

Should not be problem once we fix the previous 

> While I like the idea of directing errors out of the default location, a more lightweight approach might be a good idea. In some places, these Smarty errors appear in a red box like regular handling errors. Not sure where and why this happens, but it's certainly much less scary than the usual handling.

You can branch from 8e47b6fee88ecb739adee8ccbd4d87f78ea9ec08 and start from there

Traumflug:
> Nobody will care that the zip file is a little bit bigger

More code means less performance and people certainly care about performance.

getdatakick:
> More code means less performance and people certainly care about performance.

That's not true. These additional classes are loaded only when needed thanks to composer autoloader. So as long as debug mode is disabled, there's no performance impact whatsoever.

Traumflug:
At very least, autoloader has to search and consider more classes when additional packages exist. If performance impact is small, fine, but it's more than zero.

getdatakick:
@Traumflug could you please test the latest two commits, to see if they solved the problems you had?

getdatakick:
Because @Traumflug and @Dh42 are against adding another composer dependencies, I’ll drop the debug bar from the code, and leave only the core changes -- error handler code.

Traumflug:
Above I mentioned that some of these Smarty errors don't appear in the usual xdebug style, but as a red box. While reviewing this PR, I found one place: after clearing the cache, select back office menu -> *Catalog* -> *Products*.

For completeness of the discussion, here's a screenshot (while on branch *debug-mode*):

grafik

Traumflug:
That said, I now get this on the dashboard page as well, where I had xdebug display yesterday (see screenshot further above). On front office, error messages go to the JS console.

Looks like the new debug mode works now with xdebug as well. Excellent!

getdatakick:
These red popup boxes are the original way how error messages were handled in back office. Collected error messages are passed to smarty in `php_errors` variable https://github.com/thirtybees/thirtybees/blob/9df83b29214b62e06e554974d78f065b168aa2dc/classes/controller/AdminController.php#L4725

These error messages are rendered by [error.tpl](https://github.com/thirtybees/thirtybees/blob/1.0.x/admin-dev/themes/default/template/error.tpl) template.

It's a bit annoying, so maybe we could use the same approach as we use on frontend, and just render it to javascript library.

Traumflug:
A slight annoyance gently encourages people to work on fixing the actual issue, so I'm fine with it.

PR #829, opened 2019-01-27
Do not escape htaccess

stickyrst:
$_POST['htaccess'] should be taken as-is, and not escaped with Tools::getValue(), which strips slashes.

If .htaccess contains instructions which need backslash, saving meta's will remove those backslashes and will break the htaccess, breaking the website.


Traumflug:
Picked to _1.0.x_ to get it into the next release. Thank you very much for the contribution!

PR #845, opened 2019-02-13
Fix bad css path when shop has base URI (IE <= 9)

mcdado:
Based on https://github.com/PrestaShop/PrestaShop/pull/12554


Traumflug:
Thank you very much for the contribution!

Changed this slightly to make sure the replacement matches the beginning of the file path, only. Which led to c6b81ad47deea45c550d224fc9a8e1d306956874.

If there's something left, please don't hesitate to re-open this or open a regular issue.

PR #876, opened 2019-03-13
Getgroupsbug

chitmann:
In Multishop-Environments allways all groups are listed, even if they don't exist in a spezific Shopcontext.
The $shopCriteria variable is defined in the getGroups function, but is never used.
So added to the Query will make it possible to fix this, but you'll also have to add 'true' for the $idShop Parameter everywhere, where this function is called.
Maybe the Parameter should be true as default so it would fix it evreywhere automatically.
I can not imagine a situation where you want to have Groups in settings-fields, which aren't available in a spezific Shop.
Also the Parameter should be renamed, because it's named with id but it's only used as boolean...

### DIFF
--- a/classes/Group.php
+++ b/classes/Group.php
@@ -106,10 +106,10 @@ class GroupCore extends ObjectModel
      * @since   1.0.0
      * @version 1.0.0 Initial version
      */
-    public static function getGroups($idLang, $allShops = false)
+    public static function getGroups($idLang, $idShop = false)
     {
         $shopCriteria = '';
-        if (!$allShops) {
+        if ($idShop) {
             $shopCriteria = Shop::addSqlAssociation('group', 'g');
         }
 
--- a/classes/Group.php
+++ b/classes/Group.php
@@ -106,10 +106,10 @@ class GroupCore extends ObjectModel
      * @since   1.0.0
      * @version 1.0.0 Initial version
      */
-    public static function getGroups($idLang, $idShop = true)
+    public static function getGroups($idLang, $allShops = false)
     {
         $shopCriteria = '';
-        if ($idShop) {
+        if (!$allShops) {
             $shopCriteria = Shop::addSqlAssociation('group', 'g');
         }
 
--- a/classes/Group.php
+++ b/classes/Group.php
@@ -117,7 +117,7 @@ class GroupCore extends ObjectModel
             (new DbQuery())
                 ->select('DISTINCT g.`id_group`, g.`reduction`, g.`price_display_method`, gl.`name`')
                 ->from('group', 'g')
-                ->leftJoin('group_lang', 'gl', 'g.`id_group` = gl.`id_group` AND gl.`id_lang` = '.(int) $idLang)
+                ->leftJoin('group_lang', 'gl', 'g.`id_group` = gl.`id_group` AND gl.`id_lang` = '.(int) $idLang.$shopCriteria)
                 ->orderBy('g.`id_group` ASC')
         );
     }
--- a/classes/Group.php
+++ b/classes/Group.php
@@ -106,7 +106,7 @@ class GroupCore extends ObjectModel
      * @since   1.0.0
      * @version 1.0.0 Initial version
      */
-    public static function getGroups($idLang, $idShop = false)
+    public static function getGroups($idLang, $idShop = true)
     {
         $shopCriteria = '';
         if ($idShop) {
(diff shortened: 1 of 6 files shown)


classes/Group.php
```diff
@@ -106,18 +106,18 @@ public function __construct($id = null, $idLang = null, $idShop = null)
      * @since   1.0.0
      * @version 1.0.0 Initial version
      */
-    public static function getGroups($idLang, $idShop = false)
+    public static function getGroups($idLang, $allShops = false)
     {
         $shopCriteria = '';
-        if ($idShop) {
+        if (!$allShops) {
```
getdatakick: We can't just change semantics of input parameter. 

Call `Groups::getGroups($lang)` should return all groups regardless of shop association. That's current semantics, and it's obvious from the code that this is intended.

With your change, this call would return groups associated only with current shop. That's a big compatibility issue

I recommend to change the fix to follow original semantics.

$idShop == false --> return all groups
$idShop == true --> return groups associated with **current** shop (from context)
$idShop == number --> return groups associated with this **specific** shop

chitmann: I've reverted my changes, except the added missing use of $shopCriteria.
I've also changed the if, to add the missing behavior, if a number is given through the parameter.
Should now work as you mentioned above.

classes/Group.php
```diff
@@ -109,15 +109,17 @@ public function __construct($id = null, $idLang = null, $idShop = null)
...
+        }elseif ($idShop) {
             $shopCriteria = Shop::addSqlAssociation('group', 'g');
         }
 
         return Db::getInstance(_PS_USE_SQL_SLAVE_)->executeS(
             (new DbQuery())
                 ->select('DISTINCT g.`id_group`, g.`reduction`, g.`price_display_method`, gl.`name`')
                 ->from('group', 'g')
-                ->leftJoin('group_lang', 'gl', 'g.`id_group` = gl.`id_group` AND gl.`id_lang` = '.(int) $idLang)
+                ->leftJoin('group_lang', 'gl', 'g.`id_group` = gl.`id_group` AND gl.`id_lang` = '.(int) $idLang.$shopCriteria)
```
getdatakick: will the `->leftJoin(....)` work correctly with
```
$shopCriteria = ' INNER JOIN '._DB_PREFIX_.'group_shop group_shop ON (group_shop.id_group = g.id_group AND group_shop.id_shop = '.$idShop.')';
```
? 
I don't think so. the resulting sql would look something like this:

```sql
SELECT DISTINCT g.`id_group`, g.`reduction`, g.`price_display_method`, gl.`name`
FROM tb_group g
LEFT JOIN tb_group_lang gl ON (g.`id_group` = gl.`id_group` AND gl.`id_lang` = 1 INNER JOIN tb_group_shop group_shop ON (group_shop.id_group = g.id_group AND group_shop.id_shop = 1))
```

chitmann: Yes, it will work. 
I've dumped the resulting SQL it looks like this:

`SELECT DISTINCT g.id_group, g.reduction, g.price_display_method, gl.name FROM tb_group g LEFT JOIN tb_group_lang gl ON g.id_group = gl.id_group AND gl.id_lang = 1 INNER JOIN tb_group_shop group_shop ON (group_shop.id_group = g.id_group AND group_shop.id_shop = 2) ORDER BY g.id_group ASC`

So there are no brakets around the ON Block from the original LEFT JOIN:

`->leftJoin('group_lang', 'gl', 'g.id_group = gl.id_group AND gl.id_lang = '.(int) $idLang)`

to add the brakets it has to be:

`->leftJoin('group_lang', 'gl', '`**(**`g.id_group = gl.id_group AND gl.id_lang = '.(int) $idLang`**.')'**`)`

in this case i would have to change the other line to:

`$shopCriteria = ') INNER JOIN '._DB_PREFIX_.'group_shop group_shop ON (group_shop.id_group = g.id_group AND group_shop.id_shop = '.$idShop.')';` 

But with keeping the original line it will work this way.

BTW: 
The INNER Join is exactly the same, which you get in return from the `Shop::addSqlAssociation('group', 'g')` Method, except that i've changed the use of the  ID from context to the $idShop Parameter.

Traumflug:
This SQL query appears to work here as well, so I didn't change it. Result on branch _1.0.x_ is commit 469a569f8c98660280e3921ae277fede7d9dedc1, it'll go into the next release. Description of review edits see commit message.

Thank you for the contribution!

PR #889, opened 2019-04-01
Derive database schema from ObjectModel metadata

getdatakick:
I'm creating this pull request for review purpose.  I'd like to hear your thoughts about the design and implementation, and I'm open for any changes.

The main reason behind these changes is to make php codebase a **single source of truth** regarding database schema.  That means that thirtybees is now able to create database based solely on information stored in php metadata -- inside `ObjectModel::$definition` static property.

To achieve this I had to extend `$definitions` for all object models in the system to include additional metadata, such as database type, default value, nullable options,... Also, new kind of metadata regarding database keys / indexes was introduces.

Even though the changes is huge, there should be minimal impact on thirtybees system. 

1) changes to `$definition` -- object model definitions were mostly extended with new metadata. These changes have no impact on current system at all. However, some definition properties, such as `size`, were changed to match database. These can have *some* impact on validation. For example, thirtybees can now allow different string lengths for some fields. But the new string length is correct,  so this should be considered bugfix.

2) change to database creation - I've compared generated database agains original database using third party db comparison tool (https://github.com/DBDiff/DBDiff), and no changes were detected.


Traumflug:
Awesome work!

There's some loss of formatting, e.g. in https://github.com/thirtybees/thirtybees/pull/889/commits/31e63db0a6e08385deb0bcef641ca336c4e68417 and quite a number of missing _@since_ statements. But that's minor stuff.

I'm not sure what the point of classes as trivial as e.g. _MissingTable_ or _ExtraColumn_ is. My assumption was that such information is needed very shortly, only, and only in one place, so it could get stored on the fly, in an array during processing. Maybe I'm missing something. Maybe it's a matter of taste.

Something which would be certainly an achievement, but might not work out as well as it should, is detection of extra tables and extra columns in _DatabaseSchemaComparator_. Can't imagine how this would work out with modules also adding tables, some maybe even just adding a column to existing tables. In case this observation of mine is true, having a list of tables and columns known to have existed earlier might be a solution. This way thirty bees can at least remove its own tables/columns, but leaves those introduced by modules untouched.

In total, this looks like a really great achievement. Not guessing which changes are probably needed, but looking up what's actually there. Yay!

Dh42:
Deleting extra tables and columns would cripple a ton of shops.

getdatakick:
> I'm not sure what the point of classes as trivial as e.g. _MissingTable_ or _ExtraColumn_ is. My assumption was that such information is needed very shortly, only, and only in one place, so it could get stored on the fly, in an array during processing. Maybe I'm missing something. Maybe it's a matter of taste.

The reason why I introduced these SchemaDifferences classes is to allow database migration from one state to another. At the moment, `SchemaDifference` interface contains only `describe` used by by test. But in the near future, I want to extend this interface and add another methods to actually perform database migration. There will be methods like

- `isDestructive()`  -- returns true, if the fix would mean loosing some data. For example, dropping columns is destructive operation. Changing column type may be, but doesn't need to be (for example changing varchar(40) --> varchar(80) does not impact any data in database, the other direction might)
- `applyFix()` -- will actually perform the database migration. For example creating table, or changing type. This method is the main reason why there are so many different subclasses. Fix to define database column default value is different than fix to change column data type.
- `isNecessary` -- returns true, if the fix is required in order for thirtybees to run correctly. Basically, every Missing* is will be marked as necesseary.

With this in place, we can create migration step for CoreUpdater. Basically, we can find all fixes that needs to be executed, and fix them. We would consider only `isNecessary` changes. If all those changes are non destructive, we can run them automatically. If not, we would have to ask user what to do -- this shouldn't really happen often.  

Also, there will be new back office page that will show *all* differences, and allow user to manually fix even those that are destructive.

Traumflug:
> Also, there will be new back office page that will show all differences, and allow user to manually fix even those that are destructive.

This would be a task for Core Updater. This existing file comparison in _AdminInformationsController_ is also planned to go away, Core Updater doesn't update the file with MD5 checksums this controller uses (and having two distinct file verification procedures would be moot).

Perhaps a comment like this in each of these new classes would be a good idea:
``` php
/**
 * This class shouldn't be used by third party modules or get overriden.
 * Its interface may change without notice or deprecation period.
 */
```
This would give more freedom to adjust these classes as needed.

getdatakick:
> Perhaps a comment like this in each of these new classes would be a good idea:
> 
> ```
> /**
>  * This class shouldn't be used by third party modules or get overriden.
>  * Its interface may change without notice or deprecation period.
>  */
> ```
> 
> This would give more freedom to adjust these classes as needed.

We could also make them non-overridable. That would require some modification to PrestaShopAutoloader, but it's definitely doable. The question is, if we want to follow this direction.

If so, then these classes could be moved to Core directory... if I understand it correctly, the classes in Core directory are already not handled by PrestaShopAutoloader, but they use IoC container of sort.

Traumflug:
New day, new thoughts :-)

Is there a need for these classes in thirty bees core at all? Unless I'm missing something, they could all get moved into Core Updater. Core Updater is (will soon be) part of the distribution, so it can get used by the installer as well.

Regarding applying fixes: "asking the user" isn't really an option. If the DB schema gets changed, the developer has to decide about how destructive the change is when applying this change to the table description. Which means, there's probably no need for any investigation on "necessary" or "destructive" or whatever.

Just read today: [Cloudflare prefers Wireguard over OpenVPN/OpenSSL](https://www.tomshardware.co.uk/cloud-warp-free-vpn-service,news-60283.html), because the latter has like 600,000 lines of code, while the former works with just 4,000 lines. Keeping things simple is a bonus :-)

getdatakick:
> Is there a need for these classes in thirty bees core at all? Unless I'm missing something, they could all get moved into Core Updater. Core Updater is (will soon be) part of the distribution, so it can get used by the installer as well.

Good idea. The only other places so far that are using this new classes are installer and tests. And we can use classes from CoreUpdater from both area.

Also, added benefit is that the migration code would stay the same when we change core code.
 
> Regarding applying fixes: "asking the user" isn't really an option. If the DB schema gets changed, the developer has to decide about how destructive the change is when applying this change to the table description. Which means, there's probably no need for any investigation on "necessary" or "destructive" or whatever.

Well, that would be true if we started on green field. But we are not. There are many different installations that have many customization. Some merchants changed product name length to allow for longer names. Some dropped unique key on feature values to allow for multiple values (one of the most requested feature, btw)... And there are many more customizations. 

Thirtybees can't automatically change database structure if it would, or could, mean data loss. We can safely apply fixes like varchar(40) --> char(80), or tinyint(1) --> int(11). Or even int(11) -> int(10), as size for int type does not impact precision, only display format. 

But it much harder to justify truncating data by changing type from varchar(80) --> varchar(40), or dropping columns and tables. These (potentitally) destructive changes must be approved by merchant, always. Thankfully, most of these destructive changes are not critical. It's does not really hurt to have column varchar(80) instead of expected varchar(40). It still work.

Traumflug:
> But it much harder to justify truncating data by changing type from varchar(80) --> varchar(40), or dropping columns and tables. These (potentitally) destructive changes must be approved by merchant, always. Thankfully, most of these destructive changes are not critical. It's does not really hurt to have column varchar(80) instead of expected varchar(40). It still work.

Good point. The solution could be to avoid shortening fields. If the schema description in the class changes from varchar(30) to varchar(40), change the DB schema not always to 40 chars, but pnly if it's less than 40 chars. Problem solved.

Following typical forum discussions, such requests for decisions of merchants typically end up with posts like "didn't work", nothing else. Or merchants staying silent and not updating for 5 years. Only very few merchants remember what they did 3 years ago. Of those applying such changes, certainly 80% don't understand what they do when they follow some suggestion like "copy this text to there and it'll work". Even fewer follow thirty bees code development, which would be necessary to make an informed decision.

I see no regression or additional risk with fully automating such stuff. The PrestaShop upgrade process always applies blindly without asking. Changing this to upgrading after a close look into what's already in the DB schema is already a huge advantage. It allows intelligent automation rather than blind automation.

Dh42:
I have a ton of clients that use tables with a bunch of data that is not visible in the application. They just lengthen the product table in most cases, that way their 3rd party importers will work effectively. Some shops are going to lose a lot of data with no good solution if this is enforced.

getdatakick:
> I see no regression or additional risk with fully automating such stuff. The PrestaShop upgrade process always applies blindly without asking. Changing this to upgrading after a close look into what's already in the DB schema is already a huge advantage. It allows intelligent automation rather than blind automation.

Unfortunately there are some. While many changes are safe to apply, there are some that we can never be sure about, and that *can* (although probably wont) cause problems when fixed. I will prepare a matrix of all changes, together with impact and risk analysis, so we can talk in concrete terms.

Anyway, for any potentially destructive changes to database structure we need merchant's sign off. It's not for his/her benefit, but for ours. Nobody will be able tell that thirtybees update mechanism destroyed or corrupted his data. Thirtybees reputation could never recover from this.

For these potentially dangerous modification, we really have to show a big red warning notice to inform merchant that this particular change *could* lead to data loss or corruption. We need to explain that it is still essentia and importantl to perform the operation in order to keep system in stable and reliable state. But that the decision is up to merchant. And that he should consult with his programmer or support agency if he is not tech savvy enough to understand the decision.

Some will make database backup and apply the fix, some will ask their programmer to do it, and some will ignore it and run the thirtybees store happily with this *check engine light on*. But that's ok.

getdatakick:
I've moved schema-related classed to coreupdater module. I had to commit them to master branch, and forward submodule in order to make test and installation work. These changes is coreupdater does not impact the module functionality in any way yet, module is just a placeholder to store these classes. In the future, coreupdater will use them to migrate code.

There should be no more changes to thirtybees core required. @Traumflug , if you have no additional code review comments, we can merge this pull request

Traumflug:
Closing this, as this work was found to be awesome and got moved to Core Updater anyways. Getting it into doing something actually useful should be part of normal development, beyond this review request.

PR #892, opened 2019-04-01
BUG with Rijndael

Captain-FLAM:
no comment !


Traumflug:
Come on, this was put in intentionally to prove that more code means more bugs :-)

I mean: nice catch! Picked to 1.0.x with adjusted commit message to get it into the next release. Thank you for the contribution.

PR #897, opened 2019-04-07
Issue 871

getdatakick:
@Traumflug, could you please code review this pull request? I've tested the code and I'm pretty sure it works, but maybe I've overlooked something

### DIFF
--- a/classes/Language.php
+++ b/classes/Language.php
@@ -1281,6 +1281,10 @@ class LanguageCore extends ObjectModel
      */
     public static function getIsoById($idLang)
     {
+        if (!static::$_LANGUAGES) {
+            static::loadLanguages();
+        }
+
         if (isset(static::$_LANGUAGES[(int) $idLang]['iso_code'])) {
             return static::$_LANGUAGES[(int) $idLang]['iso_code'];
         }
--- a/classes/Mail.php
+++ b/classes/Mail.php
@@ -292,14 +292,9 @@ class MailCore extends ObjectModel
             $sendTxtContent = $configuration['PS_MAIL_TYPE'] == Mail::TYPE_BOTH || $configuration['PS_MAIL_TYPE'] == Mail::TYPE_TEXT;
             $sendHtmlContent = $configuration['PS_MAIL_TYPE'] == Mail::TYPE_BOTH || $configuration['PS_MAIL_TYPE'] == Mail::TYPE_HTML;
 
-            $moduleName = false;
+            $moduleName = static::getModuleName($templatePath, $shop);
 
-            // get templatePath
-            if (preg_match('#'.$shop->physical_uri.'modules/#', str_replace(DIRECTORY_SEPARATOR, '/', $templatePath)) && preg_match('#modules/([a-z0-9_-]+)/#ui', str_replace(DIRECTORY_SEPARATOR, '/', $templatePath), $res)) {
-                $moduleName = $res[1];
-            }
-
-            if ($moduleName !== false && (file_exists($themePath.'modules/'.$moduleName.'/mails/'.$isoTemplate.'.txt') ||
+            if ($moduleName && (file_exists($themePath.'modules/'.$moduleName.'/mails/'.$isoTemplate.'.txt') ||
                     file_exists($themePath.'modules/'.$moduleName.'/mails/'.$isoTemplate.'.html'))
             ) {
                 $templatePath = $themePath.'modules/'.$moduleName.'/mails/';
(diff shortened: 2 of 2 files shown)


Traumflug:
https://github.com/thirtybees/thirtybees/blob/issue-871/classes/Mail.php Line 135ff is inefficient:
```
        $idShop = (int)$idShop;
        if (!$idShop) {
            $idShop = (int)Context::getContext()->shop->id;
        }
[...]
        $shop = new Shop($idShop);
```
It always fetches a new _Shop_ instance, which causes at least 2 DB queries in its constructor. This should do better:
```
        $shop = Context::getContext()->shop;
        if ($idShop) {
            $shop = new Shop($idShop);
        }
```
----
https://github.com/thirtybees/thirtybees/blob/issue-871/classes/Mail.php#L779-L787
`Mail->getThemePath()` is inefficient as well: each shop instance has a theme directory, so this can melt down to a single line, making the new method obsolete:
```
        $themePath = _PS_ALL_THEMES_DIR_.$shop->directory;
```
"constants" like `_THEME_NAME_` or `_PS_THEME_DIR_` are not really constants, but kind of a global variable. Obviously a remnant of the days before there were multishops, not good OOP design. IMHO these should go away sooner or later, in favor of using class properties directly.

----
https://github.com/thirtybees/thirtybees/blob/issue-871/classes/Mail.php#L822
Why is there a custom method for logging something? Calling _Logger_ directly should do it., making code simpler, faster and just as easy to read.

----
https://github.com/thirtybees/thirtybees/blob/issue-871/classes/Mail.php#L757
Yet again a custom method just for logging/sending a simple error message in just one error case. If we'd go down this road everywhere (adding a custom method for every detectable error case), we'd end up with 2/3 of the overall code just for handling errors which should never happen anyways, likely reducing performance quite a bit. It's an interpreted language, so length of the code file matters, each line gets interpreted, each line consumes a bit memory.

While detecting and reporting error conditions is indeed the right thing to do, it should be doable in a single line, without custom methods. _Logger_ should have everything needed.

----
https://github.com/thirtybees/thirtybees/blob/issue-871/classes/Mail.php#L742
According to the method description, `$suffix` is never `'en'`. Might be a typo.

----
Whether your code works? If it fixed your use case it's pretty sure better than before.

Despite all the moaning above: thank you for the contribution!

getdatakick:
Thanks for the code review. I have fixed 
1. initialization of $shop variable
2. simplified initialization of $themePath (although there is not $shop->directory property)
3. typo with `$suffix` --> should have been `$iso`. Great catch, this might have caused infinite recursion

The other comments regarding new methods -- I'm strong believer in decomposition. Every method should do one thing only, and it should be easy to read, and easy to reason about. That's why I often extract away code that has nothing to do with the method itself -- to make it more readable. For example, compare this:

```php
    protected static function getTemplatePath($template, $suffix, $iso, Shop $shop, $baseTemplatePath)
    {
        $paths = [...];

        // return first template file in paths
        foreach ($paths as $path) {
            if (file_exists($path)) {
                return $path;
            }
        }
 
        // email template was not found, log missing template
        static::logMissingTemplate($template, $suffix, $iso, $paths);

        // If template wasn't found, let's try to fallback to english template
        if ($iso !== 'en') {
            return static::getTemplatePath($template, $suffix, 'en', $shop, $baseTemplatePath);
        }

        return null;
    }
```

with this

```php
    protected static function getTemplatePath($template, $suffix, $iso, Shop $shop, $baseTemplatePath)
    {
        $paths = [...];

        // return first template file in paths
        foreach ($paths as $path) {
            if (file_exists($path)) {
                return $path;
            }
        }
 
        // email template was not found, log missing template
        $filename = $template . $suffix;
        $localPaths = array_map(function($path) {
            return str_replace(_PS_ROOT_DIR_, '', $path);
        }, $paths);
        Logger::addLog(sprintf(
            'Email template %s for language %s not found in [%s]',
            $filename,
            $iso,
            join(', ', $localPaths)
        ), 3);

        // If template wasn't found, let's try to fallback to english template
        if ($iso !== 'en') {
            return static::getTemplatePath($template, $suffix, 'en', $shop, $baseTemplatePath);
        }

        return null;
    }
```

The first one is much more readable, your brain can comprehend it on the first glance. In the second version there is unnecessary noise, and your brain start to ask questions like what is `$localPaths`, why is there join etc... Of course, you will pretty quickly understand it as well. But there shouldn't be need for that. This is helper code that has not much to do with the method itself. And yet this *formatting* code hijacked the whole method. It made it harder to reason about.

Also, I don't buy your argument about performance impact. Yes, it is interpreted language, but it's compiled into opcode, and compiler does plenty of optimisations including code inlining. In fact, compiler can much better optimize short decoupled code than one huge code chunks.

PR #902, opened 2019-04-18
Fixing readfile for favicon

eschiendorfer:
As far as I can see the favicons are saved in the format: favicon-{id_shop}.ico.

Thats why I have logs like: PHP Warning: readfile(/path/img/favicon_1.ico): failed to open stream: No such file or directory in ...


eschiendorfer:
I will do a new PR for this...

PR #920, opened 2019-05-03
Change sorting of customer threads to last message

eschiendorfer:
The default sort by id_customer_thread is very confusing. Imagine you get an answer from an old customer, which already wrote some years before. This message will be appended and has a low id_customer_thread. So as a merchant you will not even realize, where this message is.

From a practical standpoint its clear: last message needs to be default sorting.


Traumflug:
Picked to _1.0.x_ with the message above added as commit message, see 74ee0199e8a761b8add6f6b3a6d55e59ee9d4b5d. Thank you very much for the contribution!

pk4tech:
May anyone help me i am new in this?🙂☺😣

Traumflug:
With what?

pk4tech:
About this why i am receiving this

PR #935, opened 2019-05-15
Add Image Cleaning script

Zengraph:
Check all the images and files  on the server img/p directories.
Test if they are related to something usefull, if not files are deleted.
The script stops after 3 sec and restart from there when lunched again.


Zengraph:
updated pull request .. full code for deleting files unused  and databse entries not linked to file

PR #936, opened 2019-05-15
Adding Images directories and datas cleaning

Zengraph:
Added function to check all images on the server by 3sec chunks, and test if images exist and assigned to product in database.
Clean all images products directories.


Zengraph:
To early.. have to make it better for the 2 scripts in one.

classes/ShopMaintenance.php
```diff
@@ -118,4 +125,125 @@ public static function cleanAdminControllerMessages()
...
+    
+     /**
+     * Clean image products directories 'img/p'.
+     *
+     * @since 1.0.9
+     */
+    public static function CheckAndFixImages($imagesNow,$imagesMax)
+    {
+		$starttime = time(); 
+		$path = _PS_ROOT_DIR_.'/img/p/';
```
Traumflug: There is also `_PS_PROD_IMG_DIR_`, which defined this directly.

classes/ShopMaintenance.php
```diff
@@ -118,4 +125,125 @@ public static function cleanAdminControllerMessages()
...
+			if ($key < $imagesNow) continue;
+			$image_id = intval($value['id']);
+			if (!in_array($image_id, $Image_listing, true)) {
+				$count_ko++;
+				unlink($value['path']);
+			}
+			elseif (!in_array($value['type'], $Image_types_product, true) && $value['type']!='default') {
+				$count_ko++;
+				unlink($value['path']);
+			}
```
Traumflug: I miss deleting the entire image directory here for images which no longer exist. For deleting recursively, there's `Tools::deleteDirectory()`.

classes/ShopMaintenance.php
```diff
@@ -118,4 +125,125 @@ public static function cleanAdminControllerMessages()
...
+			
+			$imagesRun = $key.'|'.$imagesMax;	
+			$imagesHere = $key;	
+			$imdigits = strlen($image_id); 
+
+			switch ($imdigits) {
+				case 1:
```
Traumflug: This can be done easier and without length limitation:
```php
$directory = implode('/', str_split($image_id));
```

Zengraph: Thank you..it is obvious, i didn't think about it as i used this piece of code since too long.

classes/ShopMaintenance.php
Zengraph: I'll implemant this corectly then, nice move !

PR #938, opened 2019-05-22
Add images laundry on maintenance tasks

Zengraph:
This is a first version of the script for products images directories maintenance.
Couldn't find the proper function to remove one product image from DB..

### DIFF
--- a/classes/ShopMaintenance.php
+++ b/classes/ShopMaintenance.php
@@ -46,9 +46,11 @@ class ShopMaintenanceCore
             static::adjustThemeHeaders();
             static::optinShop();
             static::cleanAdminControllerMessages();
+            static::cleanImagesDir('new');
 
             Configuration::updateGlobalValue('SHOP_MAINTENANCE_LAST_RUN', $now);
         }
+        else static::cleanImagesDir();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


classes/ShopMaintenance.php
```diff
@@ -126,66 +126,61 @@ public static function cleanAdminControllerMessages()
...
-			if ($imagesMax > 0 && $imagesNow > 0) static::CheckAndFixImages($imagesNow, $imagesMax); 
-		}
-	}
-
- 	public static function CheckAndFixImages($imagesNow, $imagesMax)
+    public static function cleanImagesDir($new='')
     {
-		$starttime = time(); 
+	    if (isset($new) && $new=='new') {
+		    $images_list = Db::getInstance(_PS_USE_SQL_SLAVE_)->executeS('SELECT id_image FROM '._DB_PREFIX_.'image ORDER BY id_image DESC');
```
Traumflug: To find the highest number of a database column, there's MAX(): https://dev.mysql.com/doc/refman/5.7/en/example-maximum-column.html. Shout be much less memory hungry than requesting a list with 100,000's of items.

Also, thirty bees prefers usage of DbQuery. Like
```php
$imagesMax = Db::getInstance(_PS_USE_SQL_SLAVE_)->executeS(
    (new DbQuery())
        ->select('MAX(`id_image`)')
        ->from('image')
);
```

Zengraph: changes applied on the 92d8f9a updated commit.

Zengraph:
How to finalize this enhancement module ?@getdatakick

eschiendorfer:
Hello @Zengraph I just saw this. Could you explain again, when and why this is needed? For me it's not so easy to understand, what needs to be cleaned in the image folder. Especially since ShopMaintenance::run() is called every 2 minutes.

therampagerado:
@eschiendorfer, when you delete products the pictures stay on the server. I now have a module for this cleanup from code canyon but it cleared a lot of images.

Zengraph:
"I again tried to understand the code and it's very very complex without comments. Also because of the missing {} (I am not used to it)." => What code are you talking about exactly.. I don't get it ^^

The purpose of this module is to clean the images products directory, deleting all images that have nothing to do here (there can be a lot on some old installs running import/sync scripts) and also cleaning the database by removing datas for images that are not present on the server anymore.

eschiendorfer:
OK thanks for explaining. I didn't know there can be such issues. But isn't it very risky to just delete images, that we don't know? I could imagine to offer a button (with a warning), but just running this in background can lead to unwanted outcomes imo. What if a module places some images there. Or what if the import/sync scripts are intended to place it there?

From the code itself I see that 'jpg' is hardcoded. That it surely outdated with the webp support.

 _`What code are you talking about exactly.. I don't get it ^^`_

You write code like this:
```
if (!in_array($i, $Image_listing)) Image::deleteAllImages($directory);
else foreach (glob($directory . '/*' ) as $filename) {
```
I (and core files) do it like:
```
if (!in_array($i, $Image_listing)) {
   Image::deleteAllImages($directory);
}
else {
   $filenames = glob($directory . '/*');
   foreach ($filenames as $filename) {
  }
}
```
Maybe that's just me, but my brain is messed up with the first version ^^

But ok it's anway all up to Petr, what to do here.

Zengraph:
What if a module places some images there => this not possible or the module should be recoded.
First thing that attackers try is to upload pictures that are scripts, cleaning the img products directory is really good for security and saving space.

For me brackets are nice to use if needed... otherwise what is the point ? it's not necessary to decorate the code when it's super simple to read like an "if else then", of course if you do more than one action.. then brackets are needed :)

getdatakick:
We indeed need something like this.

However, I would like to see this implemented as a work queue task, not as a static method of some random class that piggybacks on back office ajax calls.

We can schedule work queue tasks to run periodically using whatever interval we wish. We can disable it. We can track errors specific to this task. In the future, we will be able to ad-hoc trigger work queue tasks manually from back office, or run them from php cli mode (great for tasks that need more resources, like this one).

@Zengraph, can you move this functionality in a separate class that implements `WorkQueueTaskCallable` interface?

Regarding style -- it would be great if we managed to keep the programming style unified, as described in **Coding Standards** section in [Contributing.md](https://github.com/thirtybees/thirtybees/blob/main/CONTRIBUTING.md). We use  [Symfony Standards](http://symfony.com/doc/current/contributing/code/standards.html):

- variables should use `$camelCaseSyntax`
- method names should not start with uppercase
- and we use braces to indicate control structure body regardless of the number of statements it contains -- always use `{` and `}` for `if` and `else` statements. 

These are mostly guidelines, but please try to use them

Zengraph:
I don't have much time now.. so I f anybody wants to jump into and modify it in the choosen guidelines, I'll be glad.
@Zengraph, can you move this functionality in a separate class that implements WorkQueueTaskCallable interface?

Zengraph:
Where can I find informations about the "WorkQueueTaskCallable" class that you mentionned 2 weeks ago ?

eschiendorfer:
I guess checking out Class WorkQueueClientCore is needed. An example seems to be used in StockAvailable Class. 

Probably even more important is checking out: DynamicPacksSynchronizationTaskCore Class. I haven't worked with interfaces yet. Have you? Cool, that I can learn something new with tb :)

Zengraph:
No, I never worked with interfaces.. and didn't know about this synchro classes as I always used my own synch scripts lunched by crontask.

It's gonna be complicated to release this module as a service.. but having this kind of tool as command line or back office PRO tool is even better in my opinion.. as I never really liked that machines do things without telling me first and ask for confirmation.

eschiendorfer:
I also use my own cronjobys. As they do heavy stuff, they run without time limits and higher memory_limit. Not sure if this class can replace all of my stuff. But it's surely a great addition, that should be used more often in future 👍

PR #945, opened 2019-05-28
Add customer message in order page

ImkerW:
I noticed we can't view customer messages in the order page, I found this function verry usefull that is why I want to suggest bringing it back for a better communication with customers.

If the quality of the code isn't good enough please improve it. I think alot of people will be happy with this.

### DIFF
--- a/override/classes/Message.php
+++ b/override/classes/Message.php
@@ -1,0 +1,19 @@
+<?php
+
+class Message extends MessageCore {
+
+    public static function getMessagesByOrderId($idOrder, $private = false, Context $context = null) {
+        $m = parent::getMessagesByOrderId($idOrder, $private, $context);
+        $o = Db::getInstance()->executeS('
+	SELECT ct.*, m.*, e.`firstname` AS efirstname, e.`lastname` AS elastname
+	FROM `' . _DB_PREFIX_ . 'customer_thread` ct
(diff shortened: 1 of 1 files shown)


yaniv14:
You cannot commit and ask PR for override class.

Traumflug:
It's pretty trivial to move such code into the core class, so no objection.

The more interesting question is whether this code is helpful or fixes something.

therampagerado:
It is helpful for me - sometimes the customers require Pickup delivery at local depot and they ask for that in message at the end of the order. Then I have to click on Customer service to notice that. Needless to say - few times this didn't happen :D

Good addition imho!

Also - when customer enters only one phone number (home or mobile) they don't display sometimes (I belive it's home number). If this could be fixed by:

If(mobile)->display(mobile)
Ifno(mobile)->display(home)

This could be great too!

getdatakick:
This pull request seems to be irelevant now.

Since commit 386503afb00febeaf39943a8d033c3225ce6d40c, method `Message::getMessagesByOrderId` is no longer used anywhere in the core (we should deprecate it).

I'm closing this pull request, as it looks like the `CustomerMessage::getMessagesByOrderId` implements the same think already. If it's not true. If not, please open new pull request, based on latest codebase.

PR #958, opened 2019-06-04
Fix selected theme module translation

yaniv14:
Submitted form was always submitting the default theme value

### DIFF
--- a/admin-dev/themes/default/template/controllers/translations/helpers/view/translation_modules.tpl
+++ b/admin-dev/themes/default/template/controllers/translations/helpers/view/translation_modules.tpl
@@ -96,7 +96,7 @@
       <div class="panel">
         <input type="hidden" name="lang" value="{$lang}" />
         <input type="hidden" name="type" value="{$type}" />
-        <input type="hidden" name="theme" value="{$default_theme_name}" />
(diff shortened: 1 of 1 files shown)


getdatakick:
This problem seems to be injected by commit 5b7ec1ecfff0543f71dd472627d879b40a2ca128.

I'm not sure if this fix is complete, though. In the commit above, I can see more occurrences of {theme_name} being replaced by {default_theme_name), for example this one: https://github.com/thirtybees/thirtybees/commit/5b7ec1ecfff0543f71dd472627d879b40a2ca128?diff=unified#diff-3f9869df378b3d7214be73692aa0f814R107

Traumflug:
As `$default_theme_name` is no longer in use after this commit, I'm als missing removal of its definition. This should come up empty:
```none
grep --exclude-dir cache --exclude-dir old-versions --exclude-dir filemanager --exclude-dir node_modules --exclude-dir vendor --exclude-dir jquery --exclude-dir ace -rn default_theme_name
```

getdatakick:
Picked in commit f3f0f1ab2538633111fdb902a7f2a40cbb835ae6. Thanks @yaniv14 for the fix.

PR #964, opened 2019-06-07
Fix list count when using group_by

eschiendorfer:
If you change a backoffice list with group_by, the total count does fail.

To reproduce use the simple:
`public function hookActionAdminProductsListingFieldsModifier($params) {
    $params['group_by'] = 'GROUP BY a.id_product';
}`

My fix will lead to a correct _listTotal value. The if condition is for performance reason.


Traumflug:
Could reproduce the issue and found the fix to work just fine. Picked both to 1.1.x after adding the description here to the commit messages. Thank you very much!

PR #970, opened 2019-06-15
Succed  a search even with spelling errors

Lathanao:
Hi,
Here, just an implementation of the Levenshtein algorithm.
https://www.php.net/manual/en/function.levenshtein.php

The goal is to find some products with the search functionality. Even if you get some miss spelling in the search input.
That works perfectly with the vanilla ThirtyBees shop with default product.

For example, you search  "chonney" instead of "honey", search controller will be able to find the close word, here honey, and will display honey product in the result.
Tooltip_020

Some precisions about the changes:
-------------------------------------------------
a. In case there are many close words, we take the first of the list.
b. The algorithm is really fast, only 0.005 - 0.010 second to check 2000 words !
b. The algorithm could find some words, really far for the original (schlzcolad -> chocolat).
c. Later in a another commit, we could choose the better word following some requirements.

To explain a little what I added:
-----------------------------------------
1. Add a condition to check if result is empty or not. In case it's empty, try another search with the first word close to the query string (line 87-89).
2. Add method to find the closest word from the query string  (line 234 - 258)

Regards,
Lathanao

### DIFF
--- a/controllers/front/SearchController.php
+++ b/controllers/front/SearchController.php
@@ -87,10 +87,11 @@ class SearchControllerCore extends FrontController
             if(empty($searchResults))
                 $searchResults = Search::find((int) (Tools::getValue('id_lang')), self::findFirstClosestWord($query), 1, 10, 'position', 'desc', true);
             
-            foreach ($searchResults as &$product) {
-                $product['product_link'] = $this->context->link->getProductLink($product['id_product'], $product['prewrite'], $product['crewrite']);
-            }
-            Hook::exec('actionSearch', ['expr' => $query, 'total' => count($searchResults)]);
+            if (is_array($searchResults)) {
+                foreach ($searchResults as &$product) {
+                    $product['product_link'] = $this->context->link->getProductLink($product['id_product'], $product['prewrite'], $product['crewrite']);
+                }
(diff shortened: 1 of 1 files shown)


controllers/front/SearchController.php
```diff
@@ -227,4 +231,29 @@ public function setMedia()
...
+
+        foreach ( $words as &$item )
+            $item['leven'] = levenshtein($item['word'], $searchString);
+
+        $closestWord =  array_reduce($words, function($a, $b){
+            return $a['leven'] < $b['leven'] ? $a : $b;
+        }, array_shift($words))['word'];
+
+        return $closestWord;
+    }
```
getdatakick: This method could be optimized. The two iterations over $words array are unnecessary. We could do this in one iteration simply by remembering word with lowest 'level'. This would also reduce memory complexity from O(n) to O(1).

Another suggestion is to query only words that have appropriate length (not necessarily same length, but it should be close enough). It doesn't make much sense to compare words 'thirtybees' with 'the', as the difference will always be huge.

Lathanao:
Hi Getdatakick,

What's about the update ?

Regards.

getdatakick:
Picked (slightly modified) version via commit 5bf2e246b109d673197de6b84f0951e0f9a9f148. Thank you for contribution

PR #975, opened 2019-06-22
Added custom date formats in CSV import

ksimunovic:
Added custom date formats in CSV import based on language settings and predefined values for discount range from-to. This opens a new select dropdown in "Match your data" step of CSV import if one of the fields selected for columns is discount from or to date.

Custom date formats are applied exclusively to the fields of the discount.

### DIFF
--- a/admin-dev/themes/default/template/controllers/import/helpers/view/view.tpl
+++ b/admin-dev/themes/default/template/controllers/import/helpers/view/view.tpl
@@ -117,6 +117,17 @@
 					<p class="help-block">{l s='Indicate how many of the first rows of your file should be skipped when importing the data. For instance set it to 1 if the first row of your file contains headers.'}</p>
 				</div>
 			</div>
+			<div id="dateFormatContainer" class="form-group" {if !$date_formats}style="display:none"{/if}>
+				<label class="control-label col-lg-3" for="date_format">{l s='Date format'}</label>
+				<div id="selectDivDateFormat" class="col-lg-9">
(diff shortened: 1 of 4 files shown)


Traumflug:
Nice work!

Still it's probably a bit too specific to the given case:
- Adding another two options (having _Discount from (yyyy-mm-dd)_ as well as _Discount from (custom)_) raises the merchant's learning curve more than necessary ("What's the distinction between these two?"). Having just _Discount from_ should be sufficient. Same for _Discount to_, of course.
- This also removes the need to hide the custom date format selector in some situations. It should be fine to always show it, giving fewer surprises to merchants ("Where did you see this date format selector, I can't find it?" -- "Well, you have to select that over there first, ...").
- Date format selection should apply to all kind of dates, not only discount dates.
- And it should also apply to all kinds of imports, not only Products.

ksimunovic:
I reverted the commit I made before and created a new one according to the guidelines you gave. Date format is now shown all the time in the import if the import is for products, combinations or customer data because at other imports there are no dates.

Here are the fields I found which are using dates if you notice any other field I'll add it also.

PRODUCTS
reduction_from
reduction_to
date_expiration

COMBINATIONS
available_date

CUSTOMER
birthday
date_add

I know you mentioned there's no need to hide in some situations but as for e.g. in Categories, there are no date columns whatsoever it seemed to me it could be confusing to show it there so it's only showed where makes sense. If you still think it's better to show it everywhere then I could correct this easily.

Traumflug:
Great work!

Picked to both, branch _1.0.x_ and _1.1.x_.

Issue #982, opened 2019-06-30
Module uninstallation lacks a default message


Traumflug:
See also https://forum.thirtybees.com/topic/3179-11x-module-uninstall-bug/

Steps to reproduce:
- Uninstall a module not defining a confirmation message (`$this->confirmUninstall`).

Expected outcome:
- Either no confirmation dialog or a confirmation dialog with a default message.

Actual outcome:
- An empty confirmation dialog. No text, only buttons.

![](https://forum.thirtybees.com/uploads/monthly_2019_06/Capture.PNG.a5ef587c08acec1e72f28134e4656be8.PNG)

PR #983, opened 2019-06-30
Fix module uninstallation lacks of default message

yaniv14:
https://github.com/thirtybees/thirtybees/issues/982


SLiCK-303:
works great, good job....

Issue #984, opened 2019-07-01
Carrier free shipping option disables zone checkboxes


HeinCorbet:
The 'free shipping' option disables the checkboxes for zones. On saving the carrier the checkboxes are not send to the server. The carrier will not show up anymore in the back-office manual order creation form as available carrier, because no zone is defined for this carrier. On re-opening the carrier all zone checkboxes will show as deselected.
The issues caused by the function hideFees in the js/admin/carrier_wizard.js file.

Traumflug:
This "Free shipping" switch is planned to go away anyways, see #947.

The obvious workaround is to not use this switch. Setting carrier fees to zero should give the same results.

HeinCorbet:
That might be the case, but in the meantime while it is still there, it should not be broken functionality. 
See what you want to do with it. I think it will prevent some users from searching why it is not working.

Traumflug:
Thanks for the pull request, works just fine. Nevertheless I couldn't resist to simplify this with another commit, cde3145fb92ba89d681467a7857ec5f1708be16d.

This should solve the issue, please don't hesitate to re-open this issue in case it doesn't.

PR #985, opened 2019-07-01
updated the hideFees function to not disable the zone checkboxes

HeinCorbet:
Fix for issue #984 , jQuery selector for zone definition field chnaged to exclude the checkboxes fields. This will prevent the checkboxes from being disabed, so they will be submitted to the server and allowing the zone information to be created for the carrier. This is how it worked correctly before but was broken in an earlier release.


Traumflug:
Picked to _1.1.x_, thank you very much!

Issue #877, opened 2019-03-15
Saving AdminPPreferences in multistore breaks ASM


eschiendorfer:
I am using 1.0.8. If you use multistore and asm at the same time. It's very risk to save AdminPPreferences Controller. It will set the column advanced_stock_management in ps_product and ps_product to 0. This is obviously a bug.

To make things more clear, I took a video: https://streamable.com/pt69k

I dont know, if there is a general problem with multistore saving, as I already reported this one: https://github.com/thirtybees/thirtybees/issues/792

Traumflug:
Fix is available for testing in _Bleeding Edge_ -> _issue-877_ in Core Updater.

Traumflug:
Thanks for testing. Glad to see it works for you as well.

If there's anything left, please don't hesitate to re-open this issue or open another one.

PR #996, opened 2019-07-19
Fix for multistore-asm pppreferences bug

eschiendorfer:
This is fixing https://github.com/thirtybees/thirtybees/issues/877.

The reason is, that Tools::getValue('PS_ADVANCED_STOCK_MANAGEMENT') is always 0, when you have selected a specific shop. 

As the field is set 'PS_ADVANCED_STOCK_MANAGEMENT' field ist set 'visibility' => Shop::CONTEXT_ALL. It means, that we must check if the merchant is on CONTEXT_ALL.


Traumflug:
Thanks for this code, this was good inspiration. Adjusted fix should appear in a minute in _issue-877_. #877 is open for further comments, so closing here.

Issue #963, opened 2019-06-07
Shopping Cart shows "Free Shipping!" when weight has been exceeded for all available carriers


techjp:
If you add enough items to the cart that the total combined weight exceeds the maximum weight of all available carriers the cart will show "Free Shipping!" instead of "No Shipping Available":

![image](https://user-images.githubusercontent.com/7457115/59102729-ef3e8700-8967-11e9-8bfa-9dd3596908a5.png)

The same thing will happen on the "Shopping-cart Summary" page, the first page of the checkout process, if additional items are added to the shopping cart to the point that the total combined weight exceeds the maximum weight of all available shipping carriers:

![image](https://user-images.githubusercontent.com/7457115/59102754-054c4780-8968-11e9-8875-495ba0db5f5b.png)

Once the checkout process continues and reaches step 4 the user is finally told that no shipping carrier is available because the maximum weight of all available carriers has been exceeded:

![image](https://user-images.githubusercontent.com/7457115/59102818-2f056e80-8968-11e9-8c3f-6a65696e3c11.png)

When viewing the shopping cart it should not say "Free Shipping!" when in fact there are no carriers available. It should either show nothing at all about shipping or (better) it should say "No Carrier Available", "Shipping Not Available" or a similar message.

The default install of TB does not assign any weights to the sample products so this problem is not easily visible. If any of the devs would like access to my 1.0.8 test server to try this problem for themselves I am happy to provide credentials by PM either here or on the TB forum. (I am dynambee on the TB forums.)

techjp:
Should note that I just installed the Core Updater module and updated everything with the Bleeding Edge option. This issue still exists even after the most recent updates.

For what it's worth this behavior is identical in PS 1.6.1.24, the most recent version as of this comment.

techjp:
More details, tests, results, and discussion [on the TB forums](https://forum.thirtybees.com/topic/3105-shopping-cart-shows-free-shipping-when-weight-has-been-exceeded-for-all-available-carriers/).

ksimunovic:
I managed to reproduce this according to test cases on the forum but only on the Shopping-cart summary screen.

What I've found is that this bug happens as the result of JSON returned from add to cart request. While I was testing upper cart pop-up was updated correctly every time, as seen on screenshots but the page itself had problems. Also, all problems are found 

[...]

PR #1001, opened 2019-07-27
Added checkout error propagation to frontend and they are now converted to text representations

ksimunovic:
Added checkout error propagation to frontend and they are now converted to text representations, both for ajax on the checkout page and checkout page load, fixes thirtybees/thirtybees#963


Traumflug:
Picked to _issue-963_.

PR #1005, opened 2019-07-31
update for Wamp files checking on windows server

Zengraph:


Traumflug:
Picked to 1.1.x after fixing syntax errors. Thank you very much!

PR #1006, opened 2019-07-31
Update AdminInformationController for wamp

Zengraph:
Help to get a proper Json files listing on wamp server on windows and correct check.


Traumflug:
Picked to 1.1.x after fixing syntax errors. Thank you very much!

PR #1008, opened 2019-07-31
Update for TLS V 1.2 check

Zengraph:
The URL for paypal TLS testing is not working anymore, need another way of testing TLS 1.2 connection.

### DIFF
--- a/classes/ConfigurationTest.php
+++ b/classes/ConfigurationTest.php
@@ -317,25 +317,21 @@ class ConfigurationTestCore
     }
 
     /**
-     * @return bool
+     * @return 'ok' or nothing;
      *
      * @since 1.0.0
      */
     public static function testTlsv12()
     {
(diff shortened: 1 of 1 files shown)


Traumflug:
Well, thanks for the pull request. Uhm ...

- It's probably not a good idea to rely on some fancy private web page.
- Using PHP's Curl is frowned upon, Guzzle is the preferred choice.
- Changing the return value isn't a good idea either, code elsewhere depends on this.

Let me take this PR as a hint to a need for a fix. _tlstest.paypal.com_ is apparently gone (though still recommended [here](https://www.paypal.com/au/webapps/mpp/tls-http-upgrade)), but _api.paypal.com_ isn't.

Zengraph:
OK, thanks for answers and keep on using paypal.com server. I'll delete this PR

Issue #1002, opened 2019-07-27
Implement assets cache versioning system 


getdatakick:
Thirtybees comes with multiple assets (javascript, css, font,... files). These are cached by browser, or any other intermediate caching systems (cloudflare). 

At the moment, asset urls don't contain any version control parameters (?v=whatever), which makes it very hard to release new version of the assets. Merchants usually encounter weird problems like this one: https://forum.thirtybees.com/topic/3326-editing-products/

We should come up with some fool-proof system to tackle this problem. The assets should contain some unique identifier that forces browser (or intermediate caches) to invalidate when the content change. We could use

1. md5 of the asset file. This would be most precise. But it would put additional pressure on server, as files md5 checksum would have to be calculated with every request
2. unique identifier or release. For example, git commit hash. Every release would have different commit, so every release would force invalidation of *all* resources. Also, this does not put any additional stress on server, as release identifier is constant
3. thirtybees version -- this would work similarly to #2. But it would not work properly for switching between bleeding edge commits

Dh42:
What about the date? Like if the core updater just stored epoch time of the last run to the config table, then used that on these assets.

Traumflug:
CCC does something like this already, so this looks like a good strategy: https://forum.thirtybees.com/topic/3326-editing-products/?do=findComment&comment=29457

getdatakick:
I don't think CCC should be used in back office. In my experience CCC technology can very easily generate invalid javascript bundles. 

I have seen modules that contained bad javascript file with unclosed string, missing ending curly braces, or with unclosed comment. When such js file was included into CCC bundle, result was invalid javascript file, and non-functioning font-end site.

Obviously, this is the problem with individual module, and the easy fix is to uninstall such module. That might not be possible if we use CCC in the back office, though. Back office could be throwing thousands of javascript errors all over the place, rendering the whole administration unusable.

PR #1010, opened 2019-08-05
Versioning for css and js files in admin theme

Zengraph:
Reference issue: https://github.com/thirtybees/thirtybees/issues/1002

I choose Thirtybees version for forcing css and js files to reload on the browser cache whenever TB version Rlz change.
Next step will be to add git hash commit and store it properly in config/settings.inc and use it instead of TB version.

But if someone change files from the same version, using github, I presume they can also empty their browser cache alone.. it matters really when official public release is updated on a previous store.


Zengraph:
Need to update classes/controller/AdminController.php

"v=PS version" is not needed anymore

themes/javascript.tpl
```diff
@@ -51,7 +51,7 @@ var {$k} = '{$def|@addcslashes:'\''}';
 {/if}
 {if isset($js_files) && $js_files|@count}
 {foreach from=$js_files key=k item=js_uri}
-<script type="text/javascript" src="{$js_uri}"></script>
+<script type="text/javascript" src="{$js_uri}?version={$smarty.const._TB_VERSION_}"></script>
```
getdatakick: This will not work in all cases, as {$js_uri} can already contain `?` character. That would lead to incorrect URL with two `?` in it.

Anyway, I think it's not a best idea to do this in template. Better way would be to override `addJS` method in `AdminController`, and provide version string there (based on asset modification date, for example)

getdatakick:
js versioning implemented differently. Thank you for this pull request anyway, really appreciated

Issue #511, opened 2018-05-13
Count orders without invoice as well


Traumflug:
Discussed here: https://forum.thirtybees.com/topic/1720-1720/stats-do-not-show-orders-and-carts-correctly-tb-1-0-4

I fetched the patch provided there and aligned it to the newer syntax, see the one commit on branch *stats*: https://github.com/thirtybees/thirtybees/commit/26aa0dab78b2d4f01f9f3180930dcda93bfbe6ac

I think it's a matter of taste on what one wants to count. All orders recorded or only orders with an invoice ( = completed orders). As far as I can see, the methods changed are used by some backoffice statistics modules, only.

Dh42:
I am ok with adding it in.

Traumflug:
As usual when asking for votes, votes for and against it are 50:50 :-) Which leads to the idea that we need both, requiring a few additional code changes.

Dh42:
Maybe we can put this off and then add a few more features in the 1.1.x branch. I saw a couple stats posts on the forum that we might could merge into the core as well. 

https://forum.thirtybees.com/topic/1163/module-free-stats-unpaid-orders

https://forum.thirtybees.com/topic/1163/module-free-stats-unpaid-orders

Traumflug:
Probably related to this: https://forum.thirtybees.com/topic/2600-stats-module-bug-multistore/?tab=comments#comment-25801

> The issue is when you untick the option "Allow a customer to download and view PDF versions of his/her invoices." in the order status, then the stats module does not consider the order as being paid and valid.
>
> In my opinion this is a little bug as there is a specific setting "Set the order as paid." - shouldn't this be the main factor of how the stats module displays best selling products and other order related modules like the dashboard?

Traumflug:
There is a patch on branch _issue-511_, which is broken. See https://travis-ci.org/thirtybees/thirtybees/jobs/571262791

> Column 'date_add' in field list is ambiguous at line 956 in file controllers/admin/AdminStatsController.php

ksimunovic:
@Traumflug This should fix the failing Travis CI run

PR #1011, opened 2019-08-06
Update stats in order to show real infos when invoices where not gene…

Zengraph:
…rated

some uses external software like ERP or CRM for their invoices, in that case there was no Stats being generated and show on dashboard, Fixed !

ref from : https://www.prestashop.com/forums/topic/367885-dashboard-stats-are-empty-after-upgrading-to-1609/


Traumflug:
Nice finding in the forum over there.

- Could you reproduce the issue reported there?
- Looking at this thread in the PS forum, this commit fixes statistics (somewhat) for shops with invoices turned off, but breaks them for shops with invoices turned on.
- There's also a simpler solution: https://www.prestashop.com/forums/topic/367885-dashboard-stats-are-empty-after-upgrading-to-1609/?tab=comments#comment-2377645

If things got described right over there, the underlying issue is that invoice date is zero in orders handled with Orders -> Invoices -> Enable invoices turned off. Else statistics appear to work.

I think a better solution would be to generate an invoice date even with invoices turned off. Description of this switch reads _'If enabled, your customers will receive an invoice for their purchase(s).'_ It doesn't say 'don't record the time when an invoice should have been generated' One could even argue that an invoice should get generated, just not sent to the customer.

Finding the right place for a fix would include looking up all code where _PS_INVOICE_ makes a distinction, then digging further to find the code making a distinction for writing invoice date into the database. Then change that to always set the date, regardless whether an invoice gets sent to the customer.

getdatakick:
Related to issue #511

getdatakick:
Thanks for the pull request. The code is already in branch issue-511. But it wasn't decided yet if (or when) this will be included into thirtybees. There are many merchants that don't want to see numbers for unpaid numbers, for example.

PR #1020, opened 2019-08-17
Fix #365. Sort modules by displayName on AdminModulesController.

### LINKED ISSUE #365: Modules should be sorted
In the Prestashop backoffice the modules are shown sorted alphabetically.

In Thirtybees there seems to be no order at all. 

Although far from perfect alphabetical order helps at least to a certain extent to put similar modules together and to keep an overview. The random order in TB serves no purpose at all.

kublaios:


mke-tb:
We probably don't need a PR since we have an issue and a branch related to it. Let's keep the track of things there. Closing this.

PR #1026, opened 2019-08-20
Allowing usage of ObjectModel without auto_increment column

eschiendorfer:
Rarely you want to use objectModel without an auto_increment as primary column. An example: I created a module, that just adds a footer description for categories. So I want to use id_category as the primary column and of course this can't be auto_increment. 

With this modification you can just use something like:
```
$categoryGenzo` = new GenzoCategory($id_category);
$categoryGenzo->force_id = true;
$categoryGenzo->id_category = $id_category;
$categoryGenzo->save();
```
Note: it's not clever, to set $categoryGenzo->id. Cause the save() method will check if $this->id has value.


getdatakick:
Thanks for the pull request. I'm not sure that this is the right direction, though. 

There should be some extension to object model metadata to advice core how to treat primary keys of object model.

eschiendorfer:
Yeah I don't know, what is the best way to do it. In general I just tried to extend on an already existing object model. This doesn't seem to be possible right now. At least not with language and shop fields. 

I believe, that $force_id was actually made for this kind of customization.

getdatakick:
> I believe, that $force_id was actually made for this kind of customization.

It definitely was. But it makes sense for auto-increment records only. 

If we want to have object models without auto-increment pk, then it should be enforced that such model is instantiated with non-null key.

I mean

```
$rec = new MyRecord();
```

Should be allowed for auto-increment models only. It should throw exception for non-auto-increment models. For non-auto-increment models, the instantiation should always look like this:

```
$rec = new MyRecord($id);
```

and, internally, the system should check if record with $id exists, and if not, it should automatically set some internal `force_id` flag, or something like that. So, whan `$rec->save()` is called, it will know whether to create or update record.

I think this would be more user/developer friendly approach. 

1. mark (somehow) record as non-auto-increment , in `static::$definition`
2. just instantiate the record with known ID, and let the heavy lifting on core.

Do you think you could update your pull request, and implement something like that?

eschiendorfer:
I have read your comment multiple times and I slowly get your point. I believe, that I could do something like that. But probably not in the next days/few weeks. Because now christmas sales are starting, which means a lot of daily business for us :)

I will let this open and if nobody else, is doing something on this (what is very likely), I will update my PR in the future :)

eschiendorfer:
@getdatakick I have updated my PR now. I tried to implement the system as you described above. Now it's pretty simple:

You can define 'auto_increment' in objectModel Class. If you set it to false. You always have to initiate an objectModel with $id. 

If you don't do it, you will get an exception error. If you do it, everything should be handled by the core :) Please look carefully at it, as this is core objectModel it has a huge impact  ^^

cienislaw:
hmm, why not simply add additional category description field to the core? I think this could benefit all users and will be less complicated.

eschiendorfer:
@cienislaw In my opinion this are two different things. This solution is a general solution for "a lot of" use cases. 

Talking about the specific problem of footer category description, I would be fine with a core solution.

cienislaw:
you are right, I've focused only on missing functionality and didn't look into detail. I will open separate feature request.

eschiendorfer:
I close this PR as it's not cleanly finished and it's questionable, if this is the right approach. An objectModel without auto_increment is basically only needed, when we want to extend the core tables. For this we should anyway find a solution with a hook or so.

For guys that have a similair issue. Here is what I do right now: using an id_helper that is auto_increment. As an example with id_feature:

```
    public static $definition = array(
        'table'     => "my_module_feature",
        'primary'   => 'id_feature_helper',
        'multilang' => true,
        'fields' => array(
            'id_feature'      => array('type' => self::TYPE_INT, 'validate' => 'isUnsignedId', 'required' => true),
            'listing'         => array('type' => self::TYPE_BOOL),
            'link_rewrite'    => array('type' => self::TYPE_STRING, 'lang' => true, 'validate' => 'isLinkRewrite', 'required' => true),
        )
    );
```

```
   public function __construct($id_feature = null, $idLang = null) {
        $id_feature_helper = self::getIdFeatureHelper($id_feature);
        parent::__construct($id_feature_helper, $idLang);
        $this->id_feature = $id_feature;
    }
```

```
    private static function getIdFeatureHelper($id_feature) {
        $id_feature = (int)$id_feature;
        $query = new \DbQuery();
        $query->select('id_feature_helper');
        $query->from(self::$definition['table']);
        $query->where("id_feature={$id_feature}");
        return (int)\Db::getInstance()->getValue($query);
    }
```

This is pretty straightfoward and there are only two to note:
1. From speed perspective it's not really efficient
2. $object->id will be 'id_feature_helper' and not 'id_feature'.

Issue #602, opened 2018-09-19
Review jQuery


Traumflug:
thirty bees comes with no less than 604 jQuery files and most of them are likely many years old. File names indicate the usage of jQuery *1.11.0*, while the current jQuery version is *3.3.1*.

This needs a solution similar to using compos... wait, there's https://packagist.org/packages/components/jquery

Postponing to *1.1.x*, because most likely not 100% backwards compatible.

Traumflug:
Escalated to a bug, see https://github.com/thirtybees/community-theme-default/issues/79

Traumflug:
Another bug: https://github.com/thirtybees/thirtybees/pull/1034

PR #1034, opened 2019-09-02
Updating this 12 years old script

Zengraph:
the blocktopmenu was not working good on a shop, I had to click to show subcategories.. now it's fixed :)


Traumflug:
This will be solved together with the larger issue, see #602. Updating just one file is in general not a good idea, even if it works in this case.

Zengraph:
i understand jquery needs full update for the package, but it worked in this case.. so if it can help someone maybe in need of a quick solution between releases here is the note :)

getdatakick:
Thanks for the initiative, but it's not really best idea to selectively update only parts of the jquery libs

PR #1036, opened 2019-09-04
Add Availabilty filter search in back office

Zengraph:
adding a filter that is used on search.php class for giving more accurate results to customer in showing only available products for order.


Traumflug:
- `PS_SEARCH_INSTOCK` isn't used anywhere, so this switch is apparently pointless.
- `'hint'` wants a string, not an array.

getdatakick:
Thank you for this pull request. It was picked into branch issue-1088 for testing via commit aaec9a5f810e07dda04af8c02255d0340ee628c8

PR #1037, opened 2019-09-04
adding Search filter for show only available product for order

Zengraph:
add some sql condition for searching :
 if ($inStockSearch=='1') $sqltodo .= ' AND (sa.quantity > 0 OR sa.out_of_stock = 1)';

need also Patch-17 to work.. sorry for not having merge the 2 pulls requests..


getdatakick:
Thank you for this pull request. It was picked into branch issue-1088 for testing via commit aaec9a5f810e07dda04af8c02255d0340ee628c8

Issue #1038, opened 2019-09-04
Modules with Unicode main file cause 500 in admin


dvershinin:
Today we had installed a payment module, and now visiting Modules -> Payment, we get a 500.

The log:

![Screenshot from 2019-09-04 21-51-45](https://user-images.githubusercontent.com/250071/64286056-6da9c580-cf65-11e9-8b05-30bb6e989479.png)

It appears that the issue is caused by the code that tries to weed out modules with `eval` [here](https://github.com/thirtybees/thirtybees/blob/1.1.x/classes/module/Module.php#L943-L949).

That code fails to properly strip off `<?php` and `?>` tags in Unicode module files and fails to do so further with respect to Unicode files with BOM. 

The installed module file is both:

![Screenshot from 2019-09-04 21-50-19](https://user-images.githubusercontent.com/250071/64286329-f4f73900-cf65-11e9-97e9-0136211f07ce.png)

![Screenshot from 2019-09-04 21-50-01](https://user-images.githubusercontent.com/250071/64287304-f88bbf80-cf67-11e9-86f9-4dacbbaeca40.png)

dvershinin:
Please check this [PR](https://github.com/thirtybees/thirtybees/pull/1039) for the proposed fix. 
This resolved the 500 in our case, without regression.

Traumflug:
Thanks for the fix! Well, it just didn't pass the testsuite: https://travis-ci.org/thirtybees/thirtybees/jobs/588439008

Looking closer at it, it's probably not a good idea to convert the entire PHP file to some non-UTF encoding. Non-ASCII characters are entirely valid, just think of translations or currency symbols.

Accordingly I formed a similar patch, see 7df16596d208b0b6c7c879cfce33d68a1981547a It should appear in ~10 minutes in Core Updater as _Bleeding Edge_ -> _issue-1038_. Could you please test and look whether this one fixes the issue observed by you as well?

dvershinin:
@Traumflug without actually checking on my end, I think [this one](https://github.com/thirtybees/thirtybees/commit/7df16596d208b0b6c7c879cfce33d68a1981547a#diff-44d95efc0fd18f9f355779e981d580c0R945) just won't cut it for Unicode files.

`preg_replace` is not Unicode aware, it is single-byte function so it simply won't replace `<?php` in Unicode files.

Perhaps, replacing `UTF-8` with `utf-8//TRANSLIT` *might* fix the notice on my pull request. Don't have the time to check.

[...]

PR #1039, opened 2019-09-04
Fix handling Unicode PHP module files for #1038

dvershinin:
Convert module's PHP string contents to single-byte encoding, so that further substr functions will reliably strip off PHP tags.


Traumflug:
Doesn't work, see https://github.com/thirtybees/thirtybees/issues/1038#issuecomment-534131015

PR #1043, opened 2019-09-05
Fix variable name

mauroagr:
The variable name was $ecimals, but the right is $decimals

### DIFF
--- a/controllers/admin/AdminOrdersController.php
+++ b/controllers/admin/AdminOrdersController.php
@@ -2104,7 +2104,7 @@ class AdminOrdersControllerCore extends AdminController
         if ($products = Product::searchByName((int) $this->context->language->id, pSQL(Tools::getValue('product_search')))) {
             $decimals = 0;
             if ($currency->decimals) {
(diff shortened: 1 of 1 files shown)


Traumflug:
Nice catches!

Traumflug:
Looking closer into this, I'm no longer convinced of fe167b5.

- Rounding to `$decimals` makes sense only for prices to display. Which means, using the configuration value rather than the constant is better.
- `Tools::convertPrice()` rounds to database precision already, no need to do this again on the returned value.

Did this second commit fix some particular situation? It's well possible that all usages of `$decimals` in this method simply can get removed. Knowing which situation it solves would allow a better opinion on the best solution.

mauroagr:
@Traumflug The first commit update the ```$ecimals``` to ```$decimal```, i think was a wrong variable name.
The secund commit, i make thinking that database precision is the same as price precision, but reading now, i think make mistake.

Traumflug:
Fine. First commit was applied with 849810829525998c9799dab417a605de1e73914b already, thanks again for that. The other one is no longer needed.

PR #1054, opened 2019-09-14
BestSalesController: don't sort by position

getdatakick:
When orderby parameter value is 'position' then the page results in 500
error page. This is because the sql doesn't contain position field.

Url: /en/best-sales?orderby=position&orderway=asc

Default theme includes hack to handle this error. But it is ultimately
controller's responsibility to properly validate input.

### DIFF
--- a/classes/ProductSale.php
+++ b/classes/ProductSale.php
@@ -146,7 +146,7 @@ class ProductSaleCore
             ->where('p.`visibility` != \'none\'')
             ->where('EXISTS(SELECT 1 FROM `'._DB_PREFIX_.'category_product` cp JOIN `'._DB_PREFIX_.'category_group` cg ON (cp.id_category = cg.id_category AND cg.`id_group` '.(count(FrontController::getCurrentCustomerGroups()) ? 'IN ('.implode(',', FrontController::getCurrentCustomerGroups()).')' : '= 1').') WHERE cp.`id_product` = p.`id_product`)');
 
-        if ($orderBy !== 'price') {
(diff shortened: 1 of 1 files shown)


Traumflug:
I don't think it makes sense to blacklist one particular order term. Such errors happen just as well when asking to order by _'product'_, _'gauge'_, _'color'_, or whatever one might have a desire for. URLs asking to order by _'position'_ are simply broken.

The only way for a meaningful validation would be to compare against a whitelist, which would require either a second SQL query for getting this whitelist or dropping the _ORDER BY_ altogether, sorting in PHP after the query. Either should be done wherever ordering is requested, then.

getdatakick:
This is actually a little more complex. 

There is already a *whitelist* -- thirtybees supports only a very limited list of order-by [values](https://github.com/thirtybees/thirtybees/blob/1e82b376fe5b816d16b8ee3cdf6513f0933d4b3f/classes/controller/FrontController.php#L1348). If we try to pass anything else in the order-by parameter, it will be dropped / replaced by default order. 

This ultimately means that while url `/en/best-sales?orderby=whatever&orderway=asc` works, url `/en/best-sales?orderby=position&orderway=asc` throws ThirtyBeesDatabaseException -
Unknown column 'position' in 'order clause'.

At the moment, this problem is *solved* by hack in the [themes](https://github.com/thirtybees/niara/blob/2e054e062bd1f39a8294fb173b5add42298e941b/product-sort.tpl#L18). But it should not be theme responsibility -- it's core responsibility to correctly process it's input parameters. In this case it does not.

PR #1057, opened 2019-09-16
ImageType::getFormatedName - compatibility

getdatakick:
In 1.1.0 handling of image types was changed to better address
multistore / multitheme needs. Unfortunately, this introuces few
compatiblity issues.

One of them is the fact that many modules and/or themes are using
hardcoded image types like 'home_default' or 'large_default'. These
are now not resolved correctly, because the type is named 'home' and
'large'.

This is the fault of module using bad image type. But, unfortunately,
it is also very common usage.

This fix tries to resolve this situatio by trying to resolve image type
without _default suffix.

### DIFF
--- a/classes/ImageType.php
+++ b/classes/ImageType.php
@@ -221,6 +221,12 @@ class ImageTypeCore extends ObjectModel
             return $nameWithoutTheme.'_default';
         }
 
+        // if $name contains _default suffix, ie. home_default then try to resolve the name without this suffix
+        $pos = strpos($name, '_default');
+        if ($pos === strlen($name) - 8) {
+            return static::getFormatedName(substr($name, 0, $pos));
+        }
+
         // Give up searching.
         return $name;
     }


getdatakick:
Closing as the underlying issue was fixed by commit ba81aad0667658f1f083f8a4a033442bbd4526d2

getdatakick:
Reopening, as commit ba81aad0667658f1f083f8a4a033442bbd4526d2 didn't fix this issue

getdatakick:
This pull request is somewhat redundant now, but I still think it's a good to have. Let me explain what happens if some module wrongly uses hardcoded image type 'home_default': 

### Current situation:

- the html output will contain image link in a format `http://domain.com/1-home_default/rewrite.jpg`
- because home_default type does not exists, this request will be handled by PageNotFoundController
- controller will be able to re-map image type 'home_default' to 'home', and return correct image

### With these changes

- the html output will already contain correct image link: `http://domain.com/1-home/rewrite.jpg`
- if image exists, apache/nginx will return it directly without involving php

So, basically, this pull request saves some php processing time, and allows for better static assets optimization

Nobodaddy:
👍 A solution following Occam's razor!

PR #1069, opened 2019-10-02
Issue 129

ksimunovic:


controllers/admin/AdminImagesController.php
```diff
@@ -892,6 +892,22 @@ protected function regenerateNewImage($entityType, $idEntity)
...
+
+                                if (ImageManager::webpSupport()) {
+                                    ImageManager::resize(
+                                        $existingImage,
+                                        $process[$entityType].$imageObj->getExistingImgPath().'-'.stripslashes($imageType['name']).'2x.webp',
+                                        (int) $imageType['width'],
+                                        (int) $imageType['height'],
+                                        'webp'
+                                    );
+                                }
```
mke-tb: Should we maybe proceed here in the `else` block for the `if` case on line 897? IMO there is no need to generate webp version if generating `jpg` version is failed.

controllers/admin/AdminImagesController.php
```diff
@@ -892,6 +892,22 @@ protected function regenerateNewImage($entityType, $idEntity)
                                     'webp'
                                 );
                             }
+
+                            if (ImageManager::retinaSupport()) {
+                                if (!ImageManager::resize($existingImage, $process[$entityType].$imageObj->getExistingImgPath().'-'.stripslashes($imageType['name']).'2x.jpg', (int) $imageType['width'] * 2, (int) $imageType['height'] * 2)) {
```
mke-tb: Looks like the occurrances of this method are splitted into multiple lines, can we do it for this line as well? It would help with readability.

PR #1070, opened 2019-10-04
Improved handling of product unit price management

custompresta:


custompresta:
Do not forget to update/add database tables:

`ALTER TABLE tb_product ADD content_value DECIMAL(20,6) NOT NULL AFTER wholesale_price;`
`ALTER TABLE tb_product ADD content_unit VARCHAR(255) NOT NULL AFTER content_value;`
`ALTER TABLE tb_product_shop ADD content_value DECIMAL(20,6) NOT NULL AFTER wholesale_price;`
`ALTER TABLE tb_product_shop ADD content_unit VARCHAR(255) NOT NULL AFTER content_value;`

`CREATE TABLE IF NOT EXISTS `tb_product_unit` (
         `id_product_unit` int(11) unsigned NOT NULL AUTO_INCREMENT,
         `modifier` varchar(255) NOT NULL,
		 PRIMARY KEY (`id_product_unit`)
	)`

`CREATE TABLE IF NOT EXISTS `tb_product_unit_lang` (
         `id_product_unit` int(11) unsigned NOT NULL,
         `id_lang` int(11) unsigned NOT NULL,
         `content_name` varchar(255) NOT NULL,
         `display_name` varchar(255) NOT NULL,
		 PRIMARY KEY (`id_product_unit`,`id_lang`)`

getdatakick:
Thank you for this enhancement. Can you please explain what it does?

custompresta:
It will show up the content unit of the unit price as required in Germany.

DataKick <notifications@github.com> schrieb am Mo., 4. Nov. 2019, 10:37:

getdatakick:
I can see that there's also theme modification required. What will happen if this changes are used on themes that does not implement such modification (all commercial themes)?

twentybugs:
If the values are not assigned in the templates they will not show up.

getdatakick:
Thanks for confirmation. 

Unfortunately, thirtybees does not have any template inheritance, or similar method, to ship theme-related features. I'm afraid we will have to put this enhancement on ice for the time being, until such delivery vehicle is implemented. 

Thank you anyway for the contribution. Very appreciated.

custompresta:
There is no such thing required. If nobody wants to use it, just don't
place the variables in the template. But this display of the containing
unit like:

Content: 500ml (1,55€ per 100ml)

is at least required in Germany.

It should be integrated as a standard feature.

getdatakick:
If we include this feature in the *core* then people will expect this to work out of the box, everywhere. But, without changes to themes, it will not. This would work correctly on niara theme only (and only if merchants actually update the theme). 

That's not how core functionality should work.

It's probably better, and easier, to implement this as a module.

twentybugs:
I don't share your opinion about that, but anyway i can't change it.

Just think about it again if a normally general required feature should be only be implemented as a module and not as an core feature.

getdatakick:
If you can implement it in a way that does not require modification of all themes, I'd be happy to integrate it into core. If it's not possible, and themes modification are indeed necessary for this to work correctly, then it doesn't make much sense to integrate it at all. What would we say to merchants? Something like this?

>Hey, we have implemented this new cool feature into the core. But sorry, you can't use it. Not unless you edit your Panda or Warehouse theme, and adds this snippet here, here, and there... 

I don't think it's very user friendly, or generally accessible, feature. Isn't it better to create this as a module (could be community module), and say:

> Hey, here's a new module for advanced unit price management. If you need such feature, just install it into your store.

getdatakick:
And please, don't be offended by my rejection. It is not a reflection of your code at all. I actually like it a lot. 

It's just that I don't see a feasible way how to release this feature to end users, as part of a core.

twentybugs:
No worries ;)

PR #1110, opened 2019-11-18
Make order object available in tpl file.

eschiendorfer:
Allow to use order Object in tpl file. It's very helpful, if you want to improve the order_confirmation.tpl file. For example you wanna show the reference number or the order amount.


getdatakick:
Picked directly via commit aad0f1c2bb043ac6237efef7efbc24052efce3fd. Thanks!

PR #1129, opened 2020-01-02
Fix error Property Employee->lastname is empty

PrestaSafe:
Fixing error 500 on admin-dev/ajax.php file. 

Passing the Id employee in contructor.
And notifications can be displayed in backoffice head.
Capture d’écran 2020-01-02 à 18 44 37


getdatakick:
Picked via commit 24ad237bdae2c2304b6b7a7136f37f7e293fc834. Thank you for your contribution

PR #1130, opened 2020-01-03
displayAdminOrderContentOrder and displayAdminOrderTabOrder location

yaniv14:
It makes no sense for new order tab to be placed before default 'Status' and 'Documents' tabs.
especially because 'Status' tab has active state which cannot be altered by module hook.


getdatakick:
Thank you, picked in commit 1ac1237eec88bda57266beaa64fc4eb53a143b79

PR #1132, opened 2020-01-14
Show always the refund column in BO

eschiendorfer:
As we know from forum refunds/return management is an issue for many merchants:
- https://forum.thirtybees.com/topic/1134-1134/how-to-handle-returns
- https://forum.thirtybees.com/topic/3428-partial-refund-of-shipping-costs-not-possible

This is probably due to some bugs (which nobody exactly know if and where they exist). But in case of german merchants this also be due to a behaviour of tb. The shop system often considers, that a created order is paid. At least in german countries this is often untrue. It can happen, that you need to make a refund, even if the order is not yet paid.

This commit is a first step, to improve things a little bit. It shows the refund colums also if the order hasn't paid yet.

### DIFF
--- a/admin-dev/themes/default/template/controllers/orders/_product_line.tpl
+++ b/admin-dev/themes/default/template/controllers/orders/_product_line.tpl
@@ -113,26 +113,24 @@
 			{/if}
 		</td>
 	{/if}
-	{if ($order->hasBeenPaid())}
-		<td class="productQuantity text-center">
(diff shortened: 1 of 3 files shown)


getdatakick:
picked via commits 8c2a83d00e7205fd073eb032f9b59d4712772d20 and 47be813cc9d6cfb7156d9e930201ce0aa12a1c8b

Issue #1134, opened 2020-01-15
Customer sees private messages in order detail


eschiendorfer:
When you add a private comment on an order (don't send it to customer). The customer will still see the message on orders detail page.

PR #1135, opened 2020-01-15
Fix issue-1134

eschiendorfer:
This fixes: https://github.com/thirtybees/thirtybees/issues/1134

The implications of $private is IMO confusing and probably that was the reason for the bug. When you set $private = true, it means you will get all messages that are not privat.


getdatakick:
Thank you, picked in commits c926ee9bf3d249399aa795849318b75eb5ba8235 and 48720096ecc0cfef4e4ff928dc12624e3575ee88

PR #1136, opened 2020-01-15
Improve customer messages (from order detail page)

eschiendorfer:
This PR is a first step, to improve the customer messaging system. 

- It removes issues in multistore mode (don't open multiple threads). 
- It adds a select on the order detail page in the backoffice. It's now possible to update the customer message thread status from there. That's very useful as you often respond/send a message from there, but wanna close the thread right afterwards.

screen


getdatakick:
Picked by commits e3778cef9995ab2e3c553fc41c990295dc039be5 and 70d6906da22eb5c69a29d1785c11f31936fe0096. Thanks for the bug / enhancement

PR #1139, opened 2020-01-22
Fix issue-1138

eschiendorfer:
This will stop adding empty lines to csv files. I haven't tested this a lot. I don't know, if this will have any negative impact somewhere else. I would guess not.


getdatakick:
Thanks for the fix. I've picked it via commit 5dcfc8485e36fb92b9aa973886468628d83ea29e. I had to modify it a little bit to get rid of extra white spaces at the beginning of each line

PR #1147, opened 2020-02-18
Minimum amount of selected products

Zengraph:
https://forum.thirtybees.com/topic/3898-cartrules-voucher-minimum-amount-based-on-selected-productsonly/

Minimum amount for Voucher to be applied is calculed with selected products that are in the cart already, not on the total amount of the voucher.


getdatakick:
@Zengraph thanks for this enhancement. Please update also file `/tests/golden_files/db_schema.sql` to match the new sql schema -- otherwise this commit would fail travis build.

While doing so, please consider renaming the new column. `prodrestrict_minamount` is not very self-explaining. Moreover, minamount suggest it is a numerical field and not a boolean flag.

Zengraph:
"While doing so, please consider renaming the new column. prodrestrict_minamount is not very self-explaining. Moreover, minamount suggest it is a numerical field and not a boolean flag" 
OK => is "minimum_amount_product_restriction" a proper one to use ? But I think it's way too long, so maybe : "minamount_prodrestrict" is better then ?

getdatakick:
Very nice PR. Thank you, picked to 1.4 via commit 2eb6ae4ee6e1026d4e7b7f84fa3efd9fbefa1f75.

PR #1158, opened 2020-03-20
Convert SimpleXMLElement to array

martimarkov:
We check if $attributes is an array later on


getdatakick:
Picked via commit 6b6968862239aa38c9f21251f25ed01abb9cfae9

PR #1170, opened 2020-04-01
Extend Shipping Carriers options with additional cart total price range

toplakd:
Extending carrier options on tab 2-Shipping locations and costs:
- minimum-weight
- minimum-order-value
- maximum-order-value

The 3 new _carrier sql collumns should be added like the one for max_weight
- min_total
- max_total
- min_weight

Also there is en error for not showing Currency symbol on my added min_total and max_total, as my knowledge does not reach that high to make that work by myself.

### DIFF
--- a/classes/Carrier.php
+++ b/classes/Carrier.php
@@ -107,6 +107,12 @@ class CarrierCore extends ObjectModel
     public $max_height;
     /** @var int maximum package deep managed by the transporter */
     public $max_depth;
+    /** @var int minimum cart total managed by the transporter */
+    public $min_total;
+    /** @var int maximum cart total managed by the transporter */
+    public $max_total;
+    /** @var int minimum package weight managed by the transporter */
+    public $min_weight;
     /** @var int maximum package weight managed by the transporter */
     public $max_weight;
     /** @var int grade of the shipping delay (0 for longest, 9 for shortest) */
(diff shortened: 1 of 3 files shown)


toplakd:
Any thoughts about this shipping extension?
I'm using this extensions for 1 year without having single issues for customers when selecting carrier, by showing only the carrier that matches all criteria (weight based carrier + limited values on min/max order value)

getdatakick:
Yes, it's planned. There might be some minor tweaks to address corner cases, but we want to integrate this. Thanks for contribution

ALZA-Racing:
I just saw I Closed this by accident :(
How do I reopen it?

getdatakick:
Hmm, I can't reopen this as well, as the source repository is deleted.

Anyway, I pulled these changes into branch pr-1170. Will test and merge into main if everything is OK

getdatakick:
@ToplakD I have integrated this enhancemement to main in commit ef84e84d738394ecc7a313741b692139852469b0.

There were some small modification - for example I added option to choose if price threshold should be with or without tax.

Works very nicely. Thank you for this contribution

ToplakD:
Thank you.
Yes, it works superb, and with that one can really define carriers precisely.
Nice touch on tax incl/excl tab.

PR #1174, opened 2020-04-07
Unblocking modules from robots txt

seigieu:
I create it as separate pull request as I'm aware you may not accept it and tell me to override robots txt while it regenerates

However, there are modules that do work on TB, but sitemap does not work, and google says its it cannot reach path (google does not state its a robots txt problem) So merchant may tear his hair out of his head why it does not work. May not be proper solution, but may save save lots of frustration.


seigieu:
I'm sorry for mess... it does not seem to include commits that I wanted to request a pull.

PR #1175, opened 2020-04-07
Fixed wrong robots.txt rule created and blocing all controllers

seigieu:
Even in default installation 'statistics' meta entry has empty rewrite. producing `Disallow /en/` which blocks every url including main page


seigieu:
> 
> 
> This check was added (somewhat) recently in commit [bcd0048](https://github.com/seigieu/thirtybees/commit/bcd0048550aade3c51730c15624beb44e0cadedc). Maybe you have tested this on 1.1.0 , and not on bleeding edge?

Yes. 1.1.0 I was working on 2 versions at same time and fixed in one, copied to another. My bad.

PR #1176, opened 2020-04-07
There is alot of modules that acually use this path for google

seigieu:
I create it as separate pull request as I'm aware you may not accept it and tell me to override robots txt while it regenerates if my modules needs to be accessed this way

However, there are modules that do work on TB, but sitemap does not work, and google says its it cannot reach path (google does not state its a robots txt problem) So merchant may tear his hair out of his head why it does not work. May not be proper solution, but may save save lots of frustration if module is still not intended and tested on TB

Google cannot access sitemap generators, feed generators etc. Google should not be blocked from accessing modules dir


getdatakick:
Picked by commit f52cc105c331c4cb054ccc800f5619f235c45965.

PR #1200, opened 2020-06-06
Define currency column on list with type price

eschiendorfer:
On helperList() class you can define type to 'price'. This will convert a float into a price. Unfortunately it's not possible to define the currency to it. This PR change adds this "feature". 

An example:
`            
'amount' => array(
   'type'  => 'price',
   'currency' => 'id_currency'
),`


getdatakick:
Picked via commit f57e4b8c77c1a2aca903c5e16774024f6ac0317d

Issue #1203, opened 2020-06-09
Wrong product quantity on admin


uguranium:
On admin panel  Product > Quantity tab is showing wrong quantity data (Designation) for combinations. I work on it mostly whole of day to understand problem and figure out it. Sharing with you how figured out:

classes\Product.php:5634

use

```
foreach ($lang as $k => $row) {
            $new_row = array_values(array_filter($lang,function ($row) use ($combinations,$k) {
                return $row['id_product_attribute'] == $combinations[$k]['id_product_attribute'];
            }));
            $combinations[$k]['attribute_designation'] = $new_row[0]['attribute_designation'];
        }
```

instead of

```
foreach ($lang as $k => $row) {
            $combinations[$k]['attribute_designation'] = $row['attribute_designation'];
        }
```

getdatakick:
Thank your for the fix proposal.

However, we will need you to describe the problem. Your code snippet without context is not very descriptive. 

Issue description should contain: 
1) Steps to reproduce the problem
2) Expected behaviour
3) Actual behaviour

Thank you

Jorgemgm:
Hi,

This kind of issues happened me too.

When I worked with combinations and imported with csv files, I had lots of wrongs quantities.

I never understand why the error happen, because most of the time it was right.  

Meantime I upgraded my hardware (space disk).

When I saw your post, I decided make the following tests with this configuration:

Preferences => Products 
 

I didn’t check “Enable advanced stock management”, “New products use advanced stock management”, “Default warehouse on new products” and “Default pack stock management” with others configuration.


I checked:

1. Whit Enable stock management (YES), Allow ordering of out-of-stock products (YES) and Display available quantities on the product page (YES).
Outcome: Work fine.

2. Whit Enable stock management (YES), Allow ordering of out-of-stock products (YES) and Display available quantities on the product page (NO).
Outcome: Work fine.

![image](https://user-images.githubusercontent.com/38436949/84722224-80343f00-af7a-11ea-8e2b-5e108154f8b1.png)


3. Whit Enable stock management (YES), Allow ordering of out-of-stock products (NO) and Display available quantities on the product page (YES).
Outcome: Work fine.

4. Whit Enable stock management (YES), Allow ordering of out-of-stock products (NO) and Display available quantities on the product page (NO).
Outcome: Work fine.

![image](https://user-images.githubusercontent.com/38436949/84722310-bec9f980-af7a-11ea-98d6-8e5357707b5b.png)


[...]

PR #1204, opened 2020-06-09
Bugfix: Wrong product combination quantity

uguranium:
Fix wrong product quantity data on admin. If product combination remove and add other combination or if order is change its not correct way to look with predefined array keys. To be sure there need to compare each id_product_attribute with combination and lang variables.


getdatakick:
Related to issue #1203

getdatakick:
Thank you. I managed to reproduce the issue (but only with direct db changes). Your fix worked, but I have decided to use different approach using subquery, see commit 24c0464deefd2242fee9986a9c22e3a44397a650.

The result is the same - with one sql query we can replace two sql queries and expensive array merge

PR #1251, opened 2020-08-26
update statistics, do not record BOTS visits.

Zengraph:
from there : https://www.mediacom87.fr/comment-corriger-les-statistiques-de-prestashop/
Let's update statistics so so visitors are more accurate when we don't record the bot visits.


jf-viguier:
Thanks for the citation of my work @Zengraph

getdatakick:
Picked via commit a076fd31a0cbea3b83100fcc07557da3bd2473a1

PR #1252, opened 2020-09-01
Update CartRule.php

Zengraph:
Coupons listing on Customer account page is not reliable, this is fixed by filtering coupons SQL request to select only group restricted coupons that have no customers assigned


getdatakick:
picked via commit 4f1e3599ab6e11617b483a5da0587c53498770d9. Thank you

PR #1254, opened 2020-09-04
Fixing problem of dynamically created select of order states

eschiendorfer:
When you create an order in BO and change the payment method, the order state select isn't updating to the default value. The reason: This select gets updated dynamically with: 
`('#payment_module_name').replaceWith(data.view)`

That's why .change is not working...

In case you wonder, what this default order states are: it was never implemented completly, but in general it's a good idea. Here you see the hardcoded part: https://github.com/thirtybees/thirtybees/blob/1.1.x/controllers/admin/AdminOrdersController.php#L287


getdatakick:
Picked via commit 30059147faf3f9ccd5f4f5bc27981e4c7534eee2. Thank you!

PR #1255, opened 2020-09-10
Allowing the usage of $order Object in actionOrderStatusUpdate Hook.

eschiendorfer:
When you work with the hook actionOrderStatusUpdate, you often have to read or even edit the $order Object as well. So it's logically to give this object as a $param into the hook execution. 

Right now it's not possible to edit the $order object. Even when you go for "new Order($id_order)", because after the hook is executed, the order gets overwritten again.

Note: Don't forget to adjust the comment, when (version) $order was introduced.


getdatakick:
Picked via commit 4afe14fd24a9afcd73cb3f03e3f9e1c9b7af443d. Thanks for the contribution

PR #1264, opened 2020-10-16
Fix IMAP password attributes

gatasi:
Users were able to see 'IMAP password' when typing and after saving. This commit fixes the field attributes.


getdatakick:
Thank you for the PR, it's a good start. However, there is a problem with this solution -- password is lost with form re-saving.

getdatakick (changes requested):
There is a problem when form is updated -- password is lost

getdatakick:
I have extended your solution, password now survives form re-saving. Picked in commit cc2e9aea5c23c95e18610c69392534429037bcec. 

Thank you for your contribution.

PR #1291, opened 2021-03-10
Using right ids in getModuleNameById condition

yannchoquet:
AdminControllers of modules may not display in back office menu due to the condition of the getModuleNameById function which always return the first module name encountered in the map array.
If the returned module is no longer in the filesystem, the isEnabledForShops function of Module class return false


getdatakick:
Thank you. This is indeed serious bug, I will push this to the 1.2.1 hotfix release.

PR #1318, opened 2021-04-16
Improve Product.php MySQL 8 compatibility

daresh:
In MySQL 8 implicit and explicit sorting for GROUP BY is removed. This code relies on that sorting, so it should be improved by adding explicit ORDER BY. It does not break backward compatibility.

More about implicit and explicit sorting for GROUP BY removal: https://mysqlserverteam.com/removal-of-implicit-and-explicit-sorting-for-group-by/

For MySQL there is a bug in back office caused by this code - on product's Quantity tab, names of combinations are wrong, causing the shop administrator to set quantities for wrong combinations.


getdatakick:
Thank you very much

PR #1321, opened 2021-04-30
Spaces in image file names

mwrouse:
I recently made a theme that somehow keeps ending up with spaces in the image names. 

Currently, the htaccess file does not support this. 

Adding `\s` to the RewriteRules where it expects the theme name allows for this to work correctly.


getdatakick:
Thank you for the contribution

PR #1327, opened 2021-06-19
Preserve unicode chars on json encode

yaniv14:
When saving hours of operation in Store, all hours array are combined into json encoded value. this cause the unicode chars to get lost.
@getdatakick Please confirm that this is safe, because I am not a php developer.

### DIFF
--- a/controllers/admin/AdminStoresController.php
+++ b/controllers/admin/AdminStoresController.php
@@ -596,7 +596,7 @@ class AdminStoresControllerCore extends AdminController
             for ($i = 1; $i < 8; $i++) {
                 $_POST['hours'][] .= Tools::getValue('hours_'.(int) $i);
             }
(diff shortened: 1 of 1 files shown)


getdatakick:
@yaniv14 The change looks pretty safe. However, I don't understand the reason behind this change, as the default json encoding should work fine as well.  Can you describe what issues prompted you to implement this fix?

yaniv14:
This is how I get it without the proposed fix:
Screenshot from 2021-06-21 10-13-22

and this is after (which is correct):
Screenshot from 2021-06-21 10-11-24

getdatakick:
Thanks, I was able to reproduce the issue. Your solution indeed works. Picked via commit 8e7c50b3590a9894a43c6a35cd3006c776bf22b3

PR #1330, opened 2021-06-22
Fix: Query for Country::getCountriesByIdShop

moaai:
The Country_shop table cannot be filtered on a nonexistent id_lang column.
Moved filtering by id_lang to table country_lang.
This issue prevents a clean installation of the Polish DPD module.


getdatakick:
Thank you for the fix. I thought it strange that this bug was not reported before, but it looks like this method is not used by core at all. However third party modules still can, and obviously do, use it.

PR #1332, opened 2021-07-16
Fixes for NL tax rules

WBNet-Wout:
In commit eab245087f107dc3948e9161f2b7d80fbf6565c1, the Dutch VAT rate was changed from 6% to 9%, but the name of the tax rules weren't changed.

The tax rate for Croatia in the "NL Standard Rate (21%)" tax rule was also accidentally set to the 9% tax rate instead of the 21% tax rate.


getdatakick:
Thank you for your contribution

PR #1334, opened 2021-07-30
Update ParentOrderController, notification fix

Zengraph:
Update for notification that item(s) are missing when trying to renew old cart.


Zengraph:
this was an easy fix in order to have a better error message But it will be better to find another proper way, the problem si that when you want to renew a cart, it does fail and ,most of the time, don't give any indication why ... But the goal will be the show the products that are no longer available and to renew the cart Without theses products, but notified that some are missing ... Have to do another commit, with more files modified !

danieltallentire:
this code change didn't work for me - I think the order of the if statements are wrong? if it fails for any reason it goes into the first if block?
I think that the duplicate() function needs reworking to actually report an "errors" block and return a partial.

If you can write a bug / issue for this, I can probably do it properly @Zengraph

PR #1345, opened 2021-09-03
Fix redirect-loop if $auth is true in FrontController

FPDK:
Fix the authentication-guard, so the user can authenticate instead of being stuck in a redirect-loop.


getdatakick (commented):
This should not be required. 

Condition `$this->auth` exists to distinguish between controllers that require authentication and those that do not. 

Both `AuthController` and `PasswordController` have this property set to `false`.

PR #1371, opened 2021-12-17
Fix reorder bug

yaniv14:
Re order was getting null for order id, and therefore the relevant cart was wrong.


getdatakick:
Thank you. This is regression in 1.3.0, I've picked the same fix into 1.3.x branch

PR #1372, opened 2021-12-20
Adding root_dir as global smarty variable

eschiendorfer:
If you want to include a tpl file in another tpl file you need to specify the root path. Normally you can work with _PS_THEME_DIR. But in some occasions you might want to include a tpl file from a module. At least for me it's not clear how to do that easily.

Instead of adding all kind of _PS_ vars, I prefered to just add the _PS_ROOT_DIR_. This allows usage like {$root_dir}{$modules_dir}.


danieltallentire:
As far as I can find, $root_dir doesn't conflict in any of the modules that can be added from the modules list.
I applied the patch to my test instance and installed all the modules.
I then did:
`grep -rnw . -e 'root_dir'` 
against my instance.

I found a few references to a locally defined $root_dir variable in the PrestaShopAutoload.php, but as far as I can see this won't conflict as it isn't a smarty template variable.

eschiendorfer:
@danieltallentire thanks a lot for looking at it :)

getdatakick:
Thank you both!!!

PR #1378, opened 2022-01-10
Fixing product search on BO while adding an order

eschiendorfer:
A lot of merchants add orders from the BO. There is a product search functionality which with ajax, but doesn't work flawless.

The search is triggered while typing without a big delay time. If you type fast (as devs do), there is no issue. The ajax call is only made once. But if you type a bit slower, multiple ajax calls are sent. It's basically by chance, which result is rendered. It's just the last one, that comes back. It often leads to bad results (full list of products). See the video:

https://user-images.githubusercontent.com/15154932/148751030-9c0b4376-63c6-4a4c-af73-e1ffef0cbf92.mp4

Instead of adding an additonal lag, my fix focuses on the rendering. It does only render the result, that still matches the input field.

### DIFF
--- a/admin-dev/themes/default/template/controllers/orders/form.tpl
+++ b/admin-dev/themes/default/template/controllers/orders/form.tpl
@@ -44,6 +44,7 @@
 	var currency_blank = false;
   var priceDisplayPrecision = 0; {* Set in displaySummary(). *}
   var priceDatabasePrecision = {$smarty.const._TB_PRICE_DATABASE_PRECISION_};
+  var request = null;
 
 	{foreach from=$defaults_order_state key='module' item='id_order_state'}
 		defaults_order_state['{$module}'] = '{$id_order_state}';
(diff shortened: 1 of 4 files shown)


danieltallentire:
@eschiendorfer it feels like it might be better here to capture the request then use $request.abort() ?


`request = $.ajax({
                type: "POST", //TODO: Must be changed to POST
                // url etc
                success: function () {
                   //....
                },
                beforeSend: function () {
                    if (request !== null) {
                        request.abort();
                    }
                }
            });`

Then it should be possible to only have one request that comes back successfully which is the last one?

eschiendorfer:
Sorry for my late reply. I completly missed your comment. I am not a fulltime Dev and got all my knowledge by working with Prestashop/Thirtybees. I am very happy, that we have another pro Dev now :)

I have read now some theory and I think I got the concept of your proposal. I have updated my PR. Can you check again, if this looks correct to you? I believe it's working as intended. At least I saw some "cancelled requests" on browser console.

Is it correct, that I need to set request to null in "success" part?

getdatakick:
Looks good @eschiendorfer. I will test and include this fix into the mainline. Thanks

getdatakick:
Works nicely. Picked via commit 992ceb518c6f6227a9c5c4afb0283f61450b586c. Thanks you guys.

PR #1400, opened 2022-03-10
Universal fix for Admin Login Panel

deweblooper:
After Chrome/Chromium update to v99 or higher you will see on Admin Login moved items of "Stay logged in" and "I forgot my password". On other browsers layout appears correctly as before.
Fix is easy, identical for any PS or TB and no invasive for other browsers appearance of Login panel.

[https://github.com/deweblooper/thirtybees/issues/1](https://github.com/deweblooper/thirtybees/issues/1)

login_layout_chrome99


getdatakick (changes requested):
Thank you for the fix. However, the css fix should not be in `override.css` file, as that is for custom overrides. The fix should go to sass file https://github.com/thirtybees/thirtybees/blob/9e8002f786ae10e0b36187871ea70245bb748f82/admin-dev/themes/default/sass/controllers/_login.sass#L77-L81
And new css files should be generated by Sass preprocessor

getdatakick:
Picked via commit ad0915f4515800e49dcfe3ea4f13161fef09a462

Issue #1035, opened 2019-09-03
Supply order reference problem


eschiendorfer:
Since the version 1.1.0 there seems to be a bug when saving supply orders. Once the order is created and resaved the product reference is messed up.

I have created a video: https://streamable.com/1tblu

I believe that this issue was introduced with https://github.com/thirtybees/thirtybees/commit/eb847cd1d0980c839230eb7d0a05dc9b19b29c18#diff-bbf3bc3c8540ec292f3107d0e6f92080 As I tested it with 1.1.0 and only changed this file. Then it works as expeced.

getdatakick:
this commit is not part of 1.1.0 release. Is the bug in 1.1.0, or in bleeding edge?

eschiendorfer:
Oh damn. Sorry for that! The issue comes already with 1.1.0.

eschiendorfer:
Actually to be honest. With 1.1.0 I even get more problems when saving. Things like: 
```
Feld: price_te erforderlich
Feld: discount_rate erforderlich
Feld: discount_value_te erforderlich
Feld: price_with_discount_te erforderlich
Feld: tax_rate erforderlich
Feld: tax_value erforderlich
Feld: price_ti erforderlich
Feld: tax_value_with_order_discount erforderlich
Feld: price_with_order_discount_te erforderlich
```

getdatakick:
That was fixed in issue #1013

eschiendorfer:
Is it possible for you to reproduce it?

getdatakick:
Anyway, I'd bet that this issue was injected by commit 078765ff2c1f8f81501b076890f353efb60e79d8. Didn't test, though

eschiendorfer:
Ok I tested it a bit now. You are right! When I delete this `public function validateController($htmlentities = true)` then I get the same problems as in bleeding edge.

getdatakick:
Well, the current version of SupplyOrderDetail file looks ok to me. 

I think that there is something wrong with the associated controller -- it probably provide or pass non-standard data down to the object model. That's probably the reason why this object model implemented its own `validateController`

getdatakick:
The root cause of this problem is in the `ObjectModel::validateController()`. This method not only validates input values, but also **copies** data from request to current object properties.

[...]

PR #1406, opened 2022-03-28
Fixing issue 1035: reference name conflict

eschiendorfer:
I am very confident, that I finally could solve issue #1035 in a more or less clean way. 

It's problematic to use generic field name like "reference" in a form that works with multiple objects. In this case SupplyOrder & SupplyOrderDetail objects. 

In objectModel->validateController there is line: https://github.com/thirtybees/thirtybees/blob/97f3f33f3af9d972ff92745757cc34ee3ae0ef48/classes/ObjectModel.php#L1331

It will always override the $supplyOrderDetail->reference with Tools::getValue('reference'). But Tools::getValue('reference') is actually the $supplyOrder->reference. 

IMO it would be quite risky/ugly to change the generic validateController. Thats why I unset the $_POST value till the $supplyOrderDetails objects were saved and add it again, when the $supplyOrder object is saved.

Note: renaming the reference field is not really simpler. You will then need to "hack" postProcess function as well. Otherwise, $supplyOrder->reference would be empty.

### DIFF
--- a/controllers/admin/AdminSupplyOrdersController.php
+++ b/controllers/admin/AdminSupplyOrdersController.php
@@ -1193,6 +1193,9 @@ class AdminSupplyOrdersControllerCore extends AdminController
             // gets the reference
             $ref = pSQL(Tools::getValue('reference'));
 
+            // Make sure the product reference won't be overridden by the supplier reference (both columns are just called "reference")
+            unset($_POST['reference']);
+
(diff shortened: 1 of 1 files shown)


getdatakick:
@eschiendorfer I can't reproduce this issue in 1.3.0 / bleeding edge anymore. Can you please provide repro steps?

Issue #1354, opened 2021-10-24
Strange bug with old product images in Order page in BO


therampagerado:
I have 1.3-edge deployment (soon to update to 1.3 final when I have time) and today I noticed in one order that one of the products in my BO Order page product list has an old image.

The product was updated quite recently (weeks before the customer made his order) and the product image was also updated to show the new product design.

I cleared all caches and local Chrome cache but the image is still old one.
I also believe the upper image is also old but the difference there is very subtle and I'm unable to be 100% sure.

BO Orders products list view:
![image](https://user-images.githubusercontent.com/11251370/138608393-62da6c48-2ba7-4ddc-b792-04c75655ce37.png)

BO Product Images view: (notice the name is different because I did not update the alt tag, there was name change, but the product is the same, same ID, etc.)
![image](https://user-images.githubusercontent.com/11251370/138608434-109f411c-2512-4f22-9304-e8a3cecd8ca8.png)

Config:
![image](https://user-images.githubusercontent.com/11251370/138608479-feed3e80-29e4-4d72-ae0e-90e32ca29a4f.png)

PR #1415, opened 2022-04-06
Fix issue 1354 (always using same product_mini path_format)

eschiendorfer:
I am experiencing this issue https://github.com/thirtybees/thirtybees/issues/1354 since years as well. That's why I tried to fix it.

As you can read in the issue notes: the BO uses different paths. I tried to find all product_mini usages and changed the code to the following format: **product_mini_{$id_product}_{$id_shop}.jpg**.

Why not using id_product_attribute (as some parts of the code did)? Either I am getting it wrong or this makes no sense at all. product_mini is the cover image in a small cached format. There is no way, you could save the cover image related to a product_attribute. You can only save different covers for different shops. Of course one could discuss, if this option should be added, but for now it's simpler to just us the path_format with $id_shop.

Note: This PR should be tested by at least another person. Neither did I test this a lot, nor do I use a lot of product_attributes.

### DIFF
--- a/classes/Image.php
+++ b/classes/Image.php
@@ -948,6 +948,10 @@ class ImageCore extends ObjectModel
         $filesToDelete[] = _PS_TMP_IMG_DIR_.'product_'.$this->id_product.'.'.$format;
         $filesToDelete[] = _PS_TMP_IMG_DIR_.'product_mini_'.$this->id_product.'.'.$format;
 
+        foreach (Shop::getCompleteListOfShopsID() as $id_shop) {
+            $filesToDelete[] = _PS_TMP_IMG_DIR_.'product_mini_'.$this->id_product.'_'.$id_shop.'.'.$format;
+        }
+
         // perform delete
         $result = true;
         foreach ($filesToDelete as $file) {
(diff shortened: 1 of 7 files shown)


getdatakick:
I believe it is wrong to drop attribute id from `product_mini_*` images. For example, in order detail list or in cart products list we want to display correct image for combination in cart, not product cover image.

getdatakick (changes requested):
product mini should contain `id_product`, `id_attribute`, and `id_shop` identifiers

eschiendorfer:
I understand. But, what would the correct image be for a combination? Just the first one that exists? I can just tell for my shop: that would always be the same. Not a problem for me, but I believe the only clean solution would be the introduction of cover image per attribute. But that would be a huge rewrite...

getdatakick:
Let's then remove product and combination from the question. How about we replaced all occurrences of `product_mini_{product_id}*` with `image_mini_{id_image}`. After all, that's what we are trying to do here -- retrieve the miniature of specific image, and it shouldn't really matter to which product or combination this image is associated with.

While we are at that, we could also replace `product_{product_id}` with `image_{image_id}`

getdatakick:
Instead of product_mini, we now use image_mini in the core. See commit. 
Thanks for the PR.

Issue #1413, opened 2022-04-02
Life hack in Orders and manual shipping label creation


therampagerado:
![image](https://user-images.githubusercontent.com/11251370/161402033-3e26363f-f3ea-4b92-aac9-d161a9ad15a4.png)
It would be nice if we can have 'click to copy' text over the address fields in order view in admin.

Currently customer, address, phone is all rendered as one field and the approach I propose will be impossible to implement.

But it will be nice to have each field shown in different div so we can click to copy them: customer names, city, country, postal code, address (this will still need manual copying in many scenarios).

TB is used mostly by small merchandisers that don't use shipping label creation modules and make their labels by hand. It takes time and this will speed things up considerably.

Also one issue I encounter almost every day is that customers in FO are offered 2 phone fields and if they input their phone in one of them (I still don't know which one, probably home phone) it does not show under the list and you need to go up the page, enter customer's profile and copy it from the bottom of that page. If this is considered and phone is always shown in this Orders page it will be great too.

Regarding the click to copy function I believe this will do us just fine. The only thing it's lacking is indication if the click was successful (some small animation under or over the text) - https://stackoverflow.com/a/62728944 But as I said in order to implement this we will need to split the address information in different rows and make them one idea large so it's more user friendly.

mwrouse:
About the phone number thing, I see both of them show up under the shipping address, both home and mobile phone numbers. 

I get this order:
```
Name
Company
Address 1
Address 2
City, State, Zip
Country
Home Phone
Mobile Phone
```

We are still on `1.2` but I hope that wouldn't have changed. Do you really not see both phones listed in that field?


Also, I'm interested in you creating your own labels? Do you actually make your own labels or do you just have to put the address into another services website?

Do you not make any labels at all, or do you just write down the address and take it to somewhere to ship? If you're doing that, anywhere but the post office you are overpaying.

mwrouse:
Going back to the original request, I think a copy button is perfectly reasonable and would be a good feature to have, for copying the entire address.

Obviously a copy button on each line would be a little ridiculous. But moving each line to an `input` element that is disabled and stylized to not look like an input element would allow you to double click each line to select-all the text on that line. I think that might be the solution to the copying each line problem.

therampagerado:
My idea is that you need to copy name, phone, postal code and probably country if you use this feature.

The address is almost all the time not useful for full copy-paste so no need imho to have full address copy function.

[...]

PR #1416, opened 2022-04-08
Backoffice Order Address Copying

mwrouse:
Added a button in the back office for copying the entire address separated by `\n`

Double clicking each line highlights a word (browser dependent) and it will copy only what is highlighted. Triple clicking highlights the entire line (browser dependent) and again will only copy what is highlighted.

This allows you to quickly copy the entire line or just part of the line (example, just the postal code).

image

There are also now some helpful functions for copying stuff to the clipboard.

For Issue #1413

### DIFF
--- a/admin-dev/themes/default/template/controllers/orders/helpers/view/view.tpl
+++ b/admin-dev/themes/default/template/controllers/orders/helpers/view/view.tpl
@@ -718,8 +718,8 @@
                         {l s='Edit'}
                       </a>
 
-                      <div ondblclick="copyDivTextContentToClipboard(event, true);" data-placement="left" data-content="{l s='Copied!'}" data-trigger="manual">
-                      {displayAddressDetail address=$addresses.delivery newLine='</div><div ondblclick="copyDivTextContentToClipboard(event, true);" data-placement="left" data-content="Copied!" data-trigger="manual">'}
+                      <div ondblclick="copyDivTextContentToClipboardAndSelect(event, true);" data-placement="left" data-content="{l s='Copied!'}" data-trigger="manual">
+                      {displayAddressDetail address=$addresses.delivery newLine='</div><div ondblclick="copyDivTextContentToClipboardAndSelect(event, true);" data-placement="left" data-content="Copied!" data-trigger="manual">'}
                       </div>
 
                       {if $addresses.delivery->other}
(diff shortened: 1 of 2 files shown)


admin-dev/themes/default/template/controllers/orders/helpers/view/view.tpl
```diff
@@ -709,11 +709,19 @@
...
+                        {l s='Copy'}
+                      </a>
                       <a class="btn btn-default pull-right" href="?tab=AdminAddresses&amp;id_address={$addresses.delivery->id}&amp;addaddress&amp;realedit=1&amp;id_order={$order->id}&amp;address_type=1&amp;token={getAdminToken tab='AdminAddresses'}&amp;back={$smarty.server.REQUEST_URI|urlencode}">
                         <i class="icon-pencil"></i>
                         {l s='Edit'}
                       </a>
-                      {displayAddressDetail address=$addresses.delivery newLine='<br />'}
+
+                      <div ondblclick="copyDivTextContentToClipboard(event, true);" data-placement="left" data-content="{l s='Copied!'}" data-trigger="manual">
+                      {displayAddressDetail address=$addresses.delivery newLine='</div><div ondblclick="copyDivTextContentToClipboard(event, true);" data-placement="left" data-content="Copied!" data-trigger="manual">'}
```
mwrouse, line 722: How does one use the `l` function inside of something like this? Should definitely pull variations of `Copied!` from the language. I'm just not sure how.

mwrouse:
I agree that it is not ideal for everything, obviously there will always be differences in wherever you are pasting the address. 

The solution implemented here would have been for copying each line directly into something.

Maybe the solution would be to:
1. Auto-copy whatever gets highlighted when you double click
2. Auto-copy the entire line when you triple click

therampagerado:
Is there such option?

mwrouse:
Yes, absolutely. I just pushed changes, you'll need to change stuff in both `tools.js` and `view.tpl` again.

therampagerado:
This works just perfect now!

getdatakick:
Thanks guys. I have picked this via commit 05affeeeb438431cc5582d671785b9ce3b9ca0a3.

getdatakick:
Small modifications were needed

1) fallback to `document.execCommand('copy');` implementation for older browser, or for non-secured context (for example back office served via http protocol)
2) address escaping - since address is rendered inside `onclick="...textaddr..."`, we need to make sure `textaddr` does not contains any html characters, especially `"`

Otherwise, very neat.

Issue #1417, opened 2022-04-20
array_key_exits() is depreceated on objects


eschiendorfer:
My log is full with warnings like:

`PHP message: PHP Deprecated: array_key_exists(): Using array_key_exists() on objects is deprecated. Use isset() or property_exists() instead in /path/SupplyOrderDetail.php on line 316`

getdatakick:
Fixed by commit eeba3b4211adbaa9064af9ea8eb5842e1bcefd0b

PR #1418, opened 2022-04-20
Fix issue 1417 

eschiendorfer:
By using isset() instead of array_key_exits() issue https://github.com/thirtybees/thirtybees/issues/1417 should be fixed 😄


getdatakick (changes requested):
Please use `property_exists` instead of `isset`. With `isset`, the code will not work if the property contains `null` value

eschiendorfer:
I have updated the PR. Hopefully I did it correctly :) I never used property_exists before.

getdatakick:
Thank you, picked via eeba3b4211adbaa9064af9ea8eb5842e1bcefd0b

Issue #1420, opened 2022-04-20
Undefined offset when using currency on lists


eschiendorfer:
My log has quite some entries like:
`PHP Notice: Undefined offset: 1 in /path/cache/smarty/compile/7e/90/b5/7e90b51aa9ad63cd299f7c4a60d7bf3d585fcd77_0.file.list_content.tpl.php on line 328`

It happens on BO lists that are using type='price' and currency=true. An example is AdminStockMvtController.php. The issue was introduced by myself in:
https://github.com/thirtybees/thirtybees/pull/1200

getdatakick:
Fixed by #1421

PR #1421, opened 2022-04-20
Fix issue 1420

eschiendorfer:
My own PR https://github.com/thirtybees/thirtybees/pull/1200 added a new useful feature. It allows to define the currency column on a HelperList in BO. 

I missed that the core uses sometimes 'currency' => true. 

This PR fixes some notice errors: https://github.com/thirtybees/thirtybees/issues/1420

### DIFF
--- a/admin-dev/themes/default/template/helpers/list/list_content.tpl
+++ b/admin-dev/themes/default/template/helpers/list/list_content.tpl
@@ -109,7 +109,7 @@
 							{/if}
 						{/if}
 					{elseif isset($params.type) && $params.type == 'price'}
-						{if isset($params.currency)}
+						{if isset($params.currency) && is_string($params.currency)}
 							{displayPrice price=$tr.$key currency=$tr.{$params.currency}}
 						{else}
 							{displayPrice price=$tr.$key}


admin-dev/themes/default/template/helpers/list/list_content.tpl
```diff
@@ -109,7 +109,7 @@
 							{/if}
 						{/if}
 					{elseif isset($params.type) && $params.type == 'price'}
-						{if isset($params.currency)}
+						{if isset($params.currency) && is_string($params.currency)}
```
getdatakick, line 112: How about we extended the condition with a check that the `$tr` array actually contains this key? Something like

```smarty
{if isset($params.currency) && is_string($params.currency) && array_key_exists($params.currency, $tr)}
```

eschiendorfer: Very good point. I haven't tested it, but it makes a lot of sense. Why haven't you taken it?

getdatakick: I thought about it a little bit more. It probably shouldn't be responsibility of `view` to validate input. I think it would be better to do such validation in php, in `HelperCore`, `HelperList` or `HelperForm`. We can do it much easily there, maybe conditionally on debug mode settings. We can much better report on source of the problem (what controller is the culprit, etc...)

PR #1422, opened 2022-04-26
Fixing and improving icons on HelperList in BO

eschiendorfer:
Note that icons on AdminCartsController and AdminStockConfigurationController are broken. This PR fixes them. I have checked for other controllers, but icons aren't used a lot and the few others seem to do it correctly.

This PR is also a little enhancement. Currently the icons path was kind of hardcoded. The img file needed to be stored in img/admin. This is bad, if you want to use this feature with a module. I changed this behaviour with backwards compatibility. I also allow, to specify an optional width or height value.

In general I love this icon possibility. It's very user friendly in practice. It saves a lot of space and a good icon is easy to "read" for humans. So I recommend, to use it more often :)

### DIFF
--- a/admin-dev/themes/default/template/helpers/list/list_content.tpl
+++ b/admin-dev/themes/default/template/helpers/list/list_content.tpl
@@ -105,7 +105,7 @@
 							{if isset($tr[$key]['class'])}
 								<i class="{$tr[$key]['class']}"></i>
 							{else}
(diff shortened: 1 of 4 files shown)


getdatakick:
Thanks. Picked via commits bda361fee7d42aefbde1cd585bc5d362194ca6df, f64c6cb0050230fb62719c5c138553e1ebebba15, and 73a5369ad99b96de23cdc2838459c6063c5a1f83

PR #1455, opened 2022-06-17
Improving handling multiple features

eschiendorfer:
As already discussed in https://forum.thirtybees.com/topic/5686-new-feature-system-testers-needed/, this PR is a huge improvement of handling multiple features. 

It allows:
- to select multiple values by chosen
- to save new features values directly from product BO page
- to define displayable values on feature_value basis but also on product basis
- to order features values by position
- to define custom separtors and display methods in FO

It also basically removes the custom values, as there is no need for it.

The PR should be completly backward compatible. But 1 little changes might be needed in most themes in product.tpl:
Example Niara: https://github.com/thirtybees/niara/blob/ff06e92cdc290d7e29545bb66c312577e24905d4/product.tpl#L556
`{$feature.value|escape:'html':'UTF-8'} needs to become {$feature.value|escape:'UTF-8'}`

To make product position sort working, we need to execute cleansPositions() once for each id_feature. Like:
```
$features = Feature::getFeatures($this->context->language->id);
$ids_feature = array_column($features, 'id_feature');
foreach ($ids_feature as $id_feature) {
    FeatureValue::cleanPositions($id_feature);
}
```


getdatakick:
Picked via commit c54ce1ffe08329afe7fa22fd1a993c046d25753c.

I have made some slight adjustments, but nothing big.

Thank you for this great enhancement!!!

PR #1473, opened 2022-07-26
Features values - display in line

e-com-presta:
This fix causes multiple feature values to be displayed on a single line on the product page.
Zaznaczenie_219


getdatakick:
Thank you. I have picked your Smarty Cache related code via commit 310be387bc3552c9a78efdf2f5d46be8ebfb17fc. I had to make small adjustments, as some encryption implementation returns binary data that can't be readily stored inside database. 

I will not pick the other part of this PR, as it is superseded with more generic implementation that is part of #1455. That PR , in slightly modified form, will be part of 1.4.0 release.

Thank you for the contribution

PR #1489, opened 2022-08-08
AdminCustomerThreads: message status redesign

deweblooper:
From point of view of eshop employee old color of statuses are confusing: open = green, closed = red,.. 
Customer Service - view old

Expecting: green used for OK / finished, red for attention needed
Customer Service - statuses new

Customer Service - list new

And also unifying labels of statuses with buttons in thread detail
Customer Service Thread buttons


getdatakick:
Thank you for the PR. I have picked it partially via commit d32d97dae7192e4d591a6447cae7d60b62d55c34.

Additional changes: instead of displaying circles we now display badges with actual status text

deweblooper:
Yes! Your variant is much cleaner and more understandable, I like it

eschiendorfer:
Oh this was already merged... I just commented on the initial thread. My opinion would be: make the closed one in a gray color. They don't need to stay out from an ux perspective.

PR #1509, opened 2022-08-26
Implementation Crawler-Detect library

e-com-presta:
setNewConnection method from Connection class generates a mass of false data. Also, I suggest implementing the Crawler-Detect library: [https://github.com/JayBizzle/Crawler-Detect](https://github.com/JayBizzle/Crawler-Detect)


getdatakick:
Picked via commit 44b3ae0d84de81ae8ed4dd41d4f41ef286a871fd. 

Thank you for this very nice and useful enhancement!

PR #1526, opened 2022-10-10
Update AdminDuplicateUrlsController.php

Zengraph:
With big catalogs, it's better to avoid cheking all products link, so i mage a function for that, it works fast with 100K+ products website, as before it did crash the server, not enough memory and waiting long time for serveur error at the end, now it works like a charm !

### DIFF
--- a/controllers/admin/AdminDuplicateUrlsController.php
+++ b/controllers/admin/AdminDuplicateUrlsController.php
@@ -69,7 +69,7 @@ class AdminDuplicateUrlsControllerCore extends AdminController
         $languages = Language::getLanguages(true, $this->context->shop->id, true);
         foreach ($languages as $idLang) {
             $urls[(int) $idLang] = [];
-			$urls[(int) $idLang] = array_merge($urls[(int) $idLang], $this->getProductDuplicateUrls($idLang));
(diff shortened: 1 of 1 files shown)


getdatakick (changes requested):
please cast $idLang parameter to int to prevent sql injection. Just in case

Zengraph:
please cast $idLang parameter to int to prevent sql injection. Just in case
=> OK which one.. as almost all $idlang are already forced to int

getdatakick:
Thank you.

Unfortunately, this needs more work.

1) your last commit actually did not fixed anything, but introduced more problems.`$id_lang=(int)$idlang;` -- the result is stored inside new variable. Also, the input is undefined variable (case sensitivity)
2) the optimization do not work properly in multistore environment
3) and most importantly, the optimization premise is flawed. This works if, and only if, the product route is based on `rewrite`. That might not be always the case. If the rewrite url for product is based on ean13, for example, and some products shared the same ean13, this new code would not detect them. Also, by filtering products, we can no longer compare different routes url. So, for example, if some there was a clash between category and product url, we might never know.

Zengraph:
OK let it down then.. I'll use it only for personal purpose

PR #1529, opened 2022-10-20
utf8mb4_general_ci character set of database tables

e-com-presta:
No need to enforce utf8mb4_general_ci character set in CoreModels and CMSRole classes.


getdatakick:
Thank you. Picked via commit 150e7a5005ffd0faf6ca814c38c6ff50a335b7e0. I had to fix the test as well

PR #1531, opened 2022-10-21
Prestashop pseudo CSV generation bug

e-com-presta:
CSV, as name suggests, is COMMA-separated values.

### DIFF
--- a/admin-dev/themes/default/template/layout-export.tpl
+++ b/admin-dev/themes/default/template/layout-export.tpl
@@ -22,9 +22,9 @@
 *  @license    http://opensource.org/licenses/afl-3.0.php  Academic Free License (AFL 3.0)
 *  International Registered Trademark & Property of PrestaShop SA
 *}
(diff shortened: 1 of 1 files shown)


getdatakick (changes requested):
This is a compatibility change that can cause a lot of grief to merchants.

I personally know one merchant that converts exported orders csv file to xml format suitable for import to his accounting software. This change would break his process. And I'm sure there will be others that depends on current behaviour.

I suggest we introduce new configuration parameter named `Export field delimiter`, and let merchants decide instead of forcing them to use comma or semicolon. We could support all [delimiter-separated formats](https://en.wikipedia.org/wiki/Delimiter-separated_values). 

For new installations, this settings option could be comma. For existing installations, we could fallback to semicolon to be backwards compatible.

Can you extend your PR to incorporate this request?

e-com-presta:
Actually, I didn't think about modules from prestashop, which may not have a separator selection in configuration as AdminImportController has, only permanently in code semicolon as a separator.
Then surely a better solution would be to extend AdminPreferencesController with a new parameter "Export field delimiter".
I made a new PR.

getdatakick:
Superseded by #1532

PR #1533, opened 2022-10-26
Removal of gravatar from backoffice

e-com-presta:
Unnecessary ornamentation that only muddies backoffice.


getdatakick (changes requested):
We can't remove `Employee::getImage` method, as it is used in core (and potentially in modules).

For example here: https://github.com/thirtybees/thirtybees/blob/03ac96ca94b58fa40540d19aa5696befa547d88e/controllers/admin/AdminLogsController.php#L208-L212 

We have to keep this method

getdatakick:
Picked via commit 1fc7b3ebb10fbd995db02af9451c3ec075016c96

PR #1537, opened 2022-10-28
Wrong search indexing of description with words separated by a line break tag

e-com-presta:
Words separated by line break tag are connected together when sanitizing search string. As a result product is not found when searching.


getdatakick:
Picked slightly modified version via commit 33aacb10c28bb893955cc0980869dcb11f79129f

Thank you

PR #1551, opened 2022-11-19
Prevent automatically changing category friendly URL

e-com-presta:
Automatically changing category friendly URL for already indexed pages only causes SEO problems.


getdatakick:
Nice. Picked via commit 789f3cf20b41bf7a4ea97d8e348f55768067441b.
Thank you

PR #1557, opened 2022-11-23
Custom smarty date_format modifier

getdatakick:
Build in smarty `date_format` modifiers raises warnings because it internally uses deprecated `strftime` function. This PR fixes this issue by implementing custom implementation of `strftime` function and using it instead


getdatakick:
Picked via commit 99b8dfbd9d2a979109b11c72de97c317f4ebf1c5.

Small changes were required, as we still need to support PHP7.0 syntax.

PR #1571, opened 2022-11-30
Update Minify library

e-com-presta:


getdatakick:
Thank you very much.

Picked via commit 42f08dcb8bc0dd06bc10fcbe091d388cd3e949fe. 

I had to amend your commit a little bit, and include composer lock files as well, so the build process is deterministic.

Issue #1597, opened 2023-01-06
purchase_supplier_price in order_detail wrong


eschiendorfer:
As already a bit discussed here: https://forum.thirtybees.com/topic/5983-all-purchase_supplier_price-in-order_detail-table-wrong/

This column should show the price, you need to pay if you reorder the product at the supplier in your DEFAULT_CURRENCY. Instead of $price/$currencyRate it does $price*currencyRate...

getdatakick:
Code usage of this column in core clearly shows that original intent was track amount in order currency.

PR #1598, opened 2023-01-06
Fixing issue #1597

eschiendorfer:
This small PR just fixes https://github.com/thirtybees/thirtybees/issues/1597 :)

Note: That this is no backward fix. Old values aren't changed. It's not really possible IMO as the needed conversion rate is not (neccessarly) the conversion rate of the order.

### DIFF
--- a/classes/ProductSupplier.php
+++ b/classes/ProductSupplier.php
@@ -210,7 +210,7 @@ class ProductSupplierCore extends ObjectModel
         );
         if ($row && isset($row['price_te'])) {
             if ($convertedPrice) {
-                return Tools::convertPrice($row['price_te'], $row['id_currency']);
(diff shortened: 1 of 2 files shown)


getdatakick (changes requested):
I'm not 100% sure this is complete fix. 

1) The price should be stored in order currency, at least code usage suggest that.
2) `price_te` is stored in supplier currency
3) `Tools::convertPrice($priceTe, $supplierCurrency, false)` convert this to store default currency

We probably have to do Tools::convertPrice once again in `OrderDetail`, to convert store default currency to order currency.

eschiendorfer:
The PR has been updated and extended by https://github.com/thirtybees/statsmodule/pull/19

getdatakick:
Very nice, thank you. Picked via commit 96bfc25738838b6a4c20504161f6044cf2b7ede8.

Issue #1601, opened 2023-01-11
Enhancement: allow filter values in getAdminLink


eschiendorfer:
I quite often have the need of generating an adminLink, which directly filters a list. 

An example:
In the AdminProductsController list I want a column that links to AdminStockMvtController. When I click the link It should immediately set the the 'product_reference' value.

Do you like this idea too? 

The implementation is quite simple:

1. Adding an additional paramater ($filters = [])) to getAdminLink.
2. Adding a line to processFilter() function, that allows to reset the current filter.

Edit: the process seems to bit more complex for ModuleAdminController, but surely it's possible as well.

If you agree with me, that this would be helpful, I will make a PR :)

PR #1602, opened 2023-01-12
Allow list filtering by url: https://github.com/thirtybees/thirtybees/issues/1601

eschiendorfer:
This PR is a proposal, how we could handle https://github.com/thirtybees/thirtybees/issues/1601. It allows you to simply use getAdminLink() function and set up filters for a list. 

Note: im my branch there are two commits. The first one implements the new feature. The second is applying it to AdminProducts List by linking the quantity to AdminStockInstantState. 

For the second commit my code is a bit ugly:
`$style = $quantity<=0 ? 'color: #FFF' : 'color: #666';`

This would be better done by css. But I don't know which files I need to change. Obviously the second commit can also be skipped completly, if it's not wished.

### DIFF
--- a/classes/controller/AdminController.php
+++ b/classes/controller/AdminController.php
@@ -626,6 +626,11 @@ class AdminControllerCore extends Controller
 
         $prefix = $this->getCookieFilterPrefix();
 
+        // Reset current filter, if forced filter was applied
+        if (Tools::isSubmit('submitFilterForced')) {
+            $this->processResetFilters();
+        }
+
         if (isset($this->list_id)) {
             foreach ($_POST as $key => $value) {
                 if ($value === '') {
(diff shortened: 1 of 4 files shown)


classes/controller/AdminController.php
```diff
@@ -638,6 +643,12 @@ public function processFilter()
             }
 
             foreach ($_GET as $key => $value) {
+
+                // Handle forced filtering parameter by url
+                if (stripos($key, 'list_idFilter_') === 0) {
+                    $key = str_replace('list_id', $this->list_id, $key);
```
getdatakick, line 649: We should ensure that only the first occurrence of `list_id` is replaced. Just in case somebody called
```php
$link->getAdminLink($c, true, [], [ 'my_list_id' => 100 ]);
```

https://www.designcise.com/web/tutorial/how-to-replace-only-the-first-occurrence-of-a-string-in-php

getdatakick (changes requested):

getdatakick:
Merged into main branch. Thank for the contribution.

PR #1606, opened 2023-01-17
Improving customer search for multiple words search

eschiendorfer:
When you have a lot of customer entries, the customer search in BO is quite useless/messy. 

Imagine you search for "John Doe". You will get the first 50 Johns and the first 50 Does. In practice that could even mean that you don't get the combination "John Doe". More often you get the "John Doe" but it's somewhere between the other result (hard to find). 

This PR fixes this. It adds a little sort_score and limit the total results to 100 only at the end. So it will show "John Doe" as the first result and only then "John McEnroe" or "Sabrina Doe".


getdatakick:
Nice one, thanks

PR #1620, opened 2023-02-01
Update Product Class for Ecotax calculation right

Zengraph:
It is necessary to not apply discount on the product price that include ecotax, it needs to be calculated without this tax and add it again afterward.

Like this PR suggest it for PS 1.6, it is tested and it works fine. https://github.com/PrestaShop/PrestaShop/pull/22694/commits/6b1f81eae21f8eeeb82f0f6998a7124fa20e3bf9

### DIFF
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -1165,29 +1165,6 @@ class ProductCore extends ObjectModel
             $price = $productTaxCalculator->addTaxes($price);
         }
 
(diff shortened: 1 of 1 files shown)


getdatakick:
Thank you, picked via commit 4b8991b9c24558ca4f434e54fe038861b7970c49 into main branch

PR #1623, opened 2023-02-02
Improving accounting stats

eschiendorfer:
As already a bit discuessed here: https://forum.thirtybees.com/topic/5983-all-purchase_supplier_price-in-order_detail-table-wrong/ tb doesn't save all pricing information historically. 

This PR does solve this issue (at least in some ways). It adds two columns to the DB:

- shipping_cost_accounting
- payment_cost_accounting

This means, that order profits don't change, if you change your cost configs in the future. In addition this PR allows to set up absolute fees for carriers. For payment this was already there. Dashboard profit values will be calculated with the new method, but is backward compatible.

**Be careful:** I have this active for 2 weeks and it seems to work perfectly. But merchants need to update their Database by core updater module. If they forget it, they will see exceptions at very critical stages (order validation for example).

Note:

- If this gets merged, the orderprofitstats should be updated as well.
- I haven't implemented any BO edit of these new columns/values. It was also not possible before. IMO it's not important at the moment, but could be extended in the future.

### DIFF
--- a/classes/module/PaymentModule.php
+++ b/classes/module/PaymentModule.php
@@ -596,6 +596,8 @@ abstract class PaymentModuleCore extends Module
                         $orderCarrier->weight = (float) $order->getTotalWeight();
                         $orderCarrier->shipping_cost_tax_excl = (float) $order->total_shipping_tax_excl;
                         $orderCarrier->shipping_cost_tax_incl = (float) $order->total_shipping_tax_incl;
+                        $id_country = $address->id_country ?? (new Address($idAddress))->id_country;
+                        $orderCarrier->setShippingCostAccounting($carrier, $orderCarrier->shipping_cost_tax_excl, $id_country, $order->conversion_rate);
                         $orderCarrier->add();
                     }
                 }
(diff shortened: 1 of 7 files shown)


getdatakick:
Please read my reply on forum. 

I don't think this is the right approach. Reporting basically use function

`f(order, config) -> approxFee`

to calculate shipping / payment fees. What your PR does is run this function when order is created and save the result into database for each order. The only reason for that is that `config` can change over time.

Instead of saving some pseudo-random approxFee into database, I suggest we change the `config` parameter to be time-based.

We could have config table like this


| date_from | date_to | fee |
|-----------|-------- |----|
|  | 2022-01-01 | 3.0 |
| 2022-01-01 | 2022-02-01 | 3.12 |
| 2022-02-01 | 2022-03-01 | 4.0  |
| 2022-03-01 | | 4.01 |

and the function to calculate fee would use the proper row based on order date

eschiendorfer:
Thanks for your reply. I like your proposal. But how would we save this new confic values? We need a new ObjectModel for it, right?

eschiendorfer:
@getdatakick have you thought about it? I want to work on this (and on the image rewrite) next weeks, so I can finish this projects 👍

getdatakick:
I wonder why this functionality is in the core. How about we extracted it to module instead?

PR #1661, opened 2023-03-28
Message to order

e-com-presta:
Saving a message to an order with a status of "private" is rather a mistake.
Customers don't see this message in the order history and think it didn't get added. They write it again, or call the store.


therampagerado:
This is the BO order page message that merchants send out to customers?

e-com-presta:
> This is the BO order page message that merchants send out to customers?

No, this is a standard message added to a frontoffice order.

PR #1673, opened 2023-05-04
Administration all category view and search

deweblooper:
On system Administration Category page is hard to manage many categories if employee don´t know whole categories tree, when catalog contains let´s say 1270 categories ordered in 15 main categories and then each other subcategories in 3-4 sublevels.
And it is not possible to search and find any category in tree, just visible on actual level, so it is not useful at all.

This enhancement add new button to header toolbar for jump to other - new view with categories full list - new page "Categories quick list"

screen-category-tree

and also add to system new controller for displaying extra page with all catalog categories (pagination with numeber of pages viewed is also there), where employee can search and find any category in catalog like charm.
This new page also use buttons for preview selected category from FO, or straight open category editor in BO.

screen-category-quicklist

### DIFF
--- a/controllers/admin/AdminCategoriesAllController.php
+++ b/controllers/admin/AdminCategoriesAllController.php
@@ -1,0 +1,209 @@
+<?php
+/**
+ * Copyright (C) 2021 thirty bees
+ *
+ * NOTICE OF LICENSE
+ *
+ * This source file is subject to the Open Software License (OSL 3.0)
+ * that is bundled with this package in the file LICENSE.md.
+ * It is also available through the world-wide-web at this URL:
+ * https://opensource.org/licenses/osl-3.0.php
+ * If you did not receive a copy of the license and are unable to
+ * obtain it through the world-wide-web, please send an email
+ * to license@thirtybees.com so we can send you a copy immediately.
+ *
+ * @author    thirty bees <contact@thirtybees.com>
+ * @copyright 2021 thirty bees
+ * @license   Open Software License (OSL 3.0)
+ */
+
+
+
+/**
+ * Class AdminCategoriesControllerCore reworked
+ *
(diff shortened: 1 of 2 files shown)


getdatakick:
I like the idea, it's quite a nice enhancement.

I tested it, and I have few comments:

1) warning is raised: `Creation of dynamic property AdminCategoriesAllController::$_category`
2) line 126 in AdminCategoriesAllController must be some copy'n'paste error: 
```php 
if (!array_key_exists('Bad SQL query', static::$cache_lang)) {
``` 
3) permissions -- since this is new controller, non-admin employees will not have access to. Installation data should be updated to add default employees profile access. 
4) does not work correctly in multi-store environments -- lines in list are duplicated

Issues 1,2 and 4 can be easily fixed. My biggest concern is about 3. Maybe the best approach here would be to implement this view as part of existing AdminCategoriesController -- simply adjust list sql if the url contains 'quickview' or something like that. What do you think about that?

deweblooper:
@getdatakick, thank you
issue 1: warning comes from tests? or from visual check in administration? I didn´t noticed, sorry. I will try new updated 30B
issue 2: I´ve stoled/adapted method from .\controllers\admin\AdminProductsController.php
Maybe it has something with my PHP version? I am stacked on 7.4 because of my client´s server and bunch of old but useful modules used but not maintained any more.
issue 4: I am dumb :D I even didn´t tried it in multistore
issue 3: yes, it was first step where I´ve started, but after seeing those requests I wasn´t able to move on, that´s why I´ve better started an isolated file for this and tried to make it work there (I am not real programmer just empirical coder :). But really it could be much more elegant to put it to original controller, you are right

therampagerado:
@deweblooper looking forward to this improvement! ;)

PR #1674, opened 2023-05-04
Product reference code - whitespace at the beginning and end

deweblooper:
Issue:
In Administration > Catalog > Products > (any product) > [Edit]
in field "Reference code" must not be allowed to use space before first or after last character in code:
"abc123" is not same as " abc123" or "abc123 ".
This is not problem for eshop employees, but big issue for sync eshop products with third party applications where reference code is used as only key to recognize pairs.

Fix:
In ".\classes\Validate.php" make function "isReference" more precise with correct regex to avoid spaces at all.

### DIFF
--- a/classes/Validate.php
+++ b/classes/Validate.php
@@ -534,7 +534,7 @@ class ValidateCore
      */
     public static function isReference($reference)
     {
-        return (bool) preg_match(Tools::cleanNonUnicodeSupport('/^[^<>;={}]*$/u'), $reference);
+        return (bool) preg_match(Tools::cleanNonUnicodeSupport('/^[^\s<>;={}]*$/u'), $reference);
     }
 
     /**


getdatakick:
We had to revert this change.

A lot of stores is using reference codes with spaces. The same validation is used for supplier reference as well. 

With this change, updating products failed -- merchants had to change their reference code(s). What is worse, order creation with these products failed as well, because order detail contains copy of reference codes.

For backwards compatibility reasons we must support spaces in reference codes.

deweblooper:
OK, and what about modify regex?
I understand when space is in code, so this option can be "true" - you are right @getdatakick 
but I don´t believe reference need space before first character, or after last characters - this is for me "false" at all

getdatakick:
That's too risky, if there are some existing data in db that has reference with leading or trailing spaces. 

We would have to accompany such change with some data migration, and trim all references in database (product, combination, suppliers, order detail, supply orders, etc). 

If you need this, you can always implement this as a custom override

deweblooper:
I´ve found a solution:
^\s+|\s+$
I think that with this fix any store can identify bad references in it or in external sources,
which is solution more handy than ignore problem of bad references  :)
but OK, I understand honestly, nevermind, decide as you need for safety of store,
I have it implemented as override to my stores

Issue #1665, opened 2023-04-07
Field too wide


musicpanda:
A very small issue: the "Free shipping starts at" weight field on the shipping preferences page is too wide. 

![shipping](https://user-images.githubusercontent.com/13930107/230587835-8c3f687a-6b59-46e9-bcfa-ef62bbf71436.jpg)

PR #1676, opened 2023-05-12
Update AdminShippingController.php

Zengraph:
Field too wide #1665 

there is some style missing on that div input group .fixed-width-lg


Zengraph:
Field too wide #1665

Issue #1630, opened 2023-02-11
Alignment of the tracking number


Jorgemgm:
On order, when we fill in the tracking number field, it isn't aligned with the expression "Tracking number".

The field of "Tracking number" should be aligned with the beginning of the expression "Tracking number".

That way, it has more length to show the link of the tracking number.

![image](https://user-images.githubusercontent.com/38436949/218241367-4729784a-d719-4048-8bcd-f35cf1b58b9e.png)

After the update, the "Tracking number" is aligned.

![image](https://user-images.githubusercontent.com/38436949/218241379-dfc9d578-3874-4129-94cb-7b01d104dc1c.png)

PR #1677, opened 2023-05-12
TO DELETE - Update _shipping.tpl

Zengraph:
TO DELETE - Alignment of the tracking number #1630


Zengraph:
Sorry it was not complete there is another one better.

PR #1678, opened 2023-05-12
Update _shipping.tpl

Zengraph:
Alignment of the tracking number #1630 
with style added.


Zengraph:
Alignment of the tracking number #1630

PR #1686, opened 2023-06-02
Pagination: wrong canonical URL

e-com-presta:
Quote: "Don't use the first page of a paginated sequence as the canonical page."
https://developers.google.com/search/docs/specialty/ecommerce/pagination-and-incremental-page-loading

### DIFF
--- a/classes/controller/FrontController.php
+++ b/classes/controller/FrontController.php
@@ -445,6 +445,10 @@ class FrontControllerCore extends Controller
                 break;
 
         }
+        // append page number
+        if ($p = Tools::getIntValue('p')) {
+            $canonical .= "?p=$p";
+        }
         // build new content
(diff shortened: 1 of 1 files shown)


classes/controller/FrontController.php
```diff
@@ -445,6 +445,10 @@ public function getSeoFields()
                 break;
 
         }
+        // append page number
+        if ($p = Tools::getIntValue('p')) {
+            $canonical .= "?p=$p";
```
getdatakick, line 450: This is not the right way to modify url.

At this point, we don't know what `$canonical` looks like. It could be `https://www.domain.com/en/rewrite/url`
but it could very easily be `https://www.domain.con/en/rewrite?id_category=10`

We can't simply concat `'?p=1` here

getdatakick (changes requested):

Issue #1475, opened 2022-07-29
ps_cart_product missing date_upd column


eschiendorfer:
The table ps_cart_product should be extended with date_upd column. It's a bad behaviour to just have date_add and even overwrite this it when actually an update happened:
https://github.com/thirtybees/thirtybees/blob/c3a495ae6bbbfb1d1c7ef4f66d03d544b34dbcc0/classes/Cart.php#L3269

Why is it important? IMO the cart should be rendered in the order, that the products were added. I believe blockcart module does handle this somehow. But when you want to use the core functions you will walk into this issue.

eschiendorfer:
Puh, I am working on my new theme and this report is almost a year old. This "issue" is basically also in core:

[screen-capture.webm](https://github.com/thirtybees/thirtybees/assets/15154932/d036be2e-5d31-410a-8e7d-4316902cc2ba)

Can I just submit a PR with a new column date_upd. Basically date_upd would be handled as date_add now and date_add is only set when the add_to_cart really happens. Or am I missing something and this is more complex?

eschiendorfer:
I already opened a PR. The default value is actually my question. How can this be set to date_add?

getdatakick:
it can't, at least not without db triggers. And that's not something we want to add right now. I thing that current_timestamp is good enough for now

PR #1699, opened 2023-07-07
Adding date_upd column to cart_product table

eschiendorfer:
This PR should solve: https://github.com/thirtybees/thirtybees/issues/1475

IMO the fix should be a bit modified. Now it's creating the column date_upd with values '0000-00-00 00:00:00'. It would actually be much better if it copies the values from date_add column. Is this possible somehow?

Otherwise the change in setNoMultishipping() might be risky:
https://github.com/thirtybees/thirtybees/compare/main...eschiendorfer:thirtybees:addDateUpdToCartProduct#diff-a1349548e1b4f0f9e9bda744305bba472b6cb9d887572fa1bcb9498f401e5fb7L4882-L4883

### DIFF
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -4882,7 +4882,7 @@ class CartCore extends ObjectModel
 						AND (cp1.id_product = cp2.id_product)
 						AND (cp1.id_product_attribute = cp2.id_product_attribute)
 						AND (cp1.id_address_delivery <> cp2.id_address_delivery)
-						AND (cp1.date_upd > cp2.date_upd)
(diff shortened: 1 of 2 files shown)


eschiendorfer:
Ok I saw now your comment about the 'required' thing. Which makes sense. I have tried to use a default value and change the code, so that it works if there is basically no useful date_upd value.

I havent tested setNoMultishipping(). You think that will work like that?

getdatakick:
Picked via commit 3da1bf5be46ed1b262140b598b6f1840b8f0bdf0.

I have changed default value to use `CURRENT_TIMESTAMP` instead of `0000-00-00 00:00:00`, and also generated sql golden file for test.

PR #1700, opened 2023-07-16
Add access to OrderCartRule via WebService

8191:
Adds an endpoint for OrderCartRule to the WS accessible via order_cart_rules


getdatakick:
Merged. Thank you!

Issue #856, opened 2019-02-24
In transit


Jorgemgm:
Localization => Translations => MODIFY TRANSLATIONS => Email templates translations => Modify => Core emails => in_transit

The in_transit email template content “You can track your package using the following link: {followup}”, but in Email received for the customer didn’t display the {followup} link.

[In transit.pdf](https://github.com/thirtybees/thirtybees/files/2898049/In.transit.pdf)

yaniv14:
@Jorgemgm Did you setup a tracking link in carrier settings?

Jorgemgm:
@ yaniv14 You are right. I didn't setup a tracking link in carrier settings.

I think TB can close this issue.

Thanks

Traumflug:
Looking at the PDF, there are other placeholders not replaced. Is this a custom message or the default one?

Also, if there's no tracking link available, the entire line should get dropped.

Jorgemgm:
@Traumflug It's default (I tested in two site tests without changes).

Jorgemgm:
The in_transit email it sends after “Tracking number” update on Orders => Orders => View => Shipping => Tracking number => Edit => Update

We think it’s good idea, on email, beside track link include the “Tracking number” {shipping_number}.

getdatakick:
The `{shipping_number}` placeholder is already passed to email template

Issue #297, opened 2017-07-06
Track & Trace Variable for e-mail tremplate


benzon:
Think an important change that needs to happen quite fast, is actually adding Track and Trace variable for the e-mail templates where it could be used and would make sence, so we can actually send out the track and trace link, and shipping number when the costumer gets the mail when you changes the status to Shipped.

Important feature in my view.

firstred:
Don't forget country iso, postcode, numerical part of mixed postcodes etc.
We will likely need to patch in a lot of stuff. I think it's okay starting with 1.0.2 despite it being a patch version, PS' implementation just sucked.

benzon:
If i knew how to do the class overwrights i would fix it that way currently, but ye im not that in to doing modules yet, since prestashop coding is fairly new.

But i agree, there is some things that needs to be changed, the reason i just thaught of it is our carrier again again got issues sending out track and trace to costumers, and we just want to make sure they get it.

benzon:
You just gotta "love" gls HAHA :D

firstred:
Doesn't gls just use one tracking number?

Are you using the `in_transit` mail?

benzon:
Quick question i know this is dumb, the translations for the Status of the order, Like Payment error, Backorder and so on where are they located? i cant find then currently in the translation part.

firstred:
Orders > Statuses

benzon:
Perfect, but still would make sence to use the shipped template to for adding the track and trace?

Since in_transit and shipped more or less do the same thing in the end in my point of view.

firstred:
Not if you have multiple shipments per order.

firstred:
Postponed till 1.0.3 though. 1.0.2 is almost ready to be shipped.

benzon:
How and when is the in_transit mail send out? We use a module for handling shipping GLS module for Prestashop, and im not sure when the in_transit is triggered and how.

benzon:
Tryed digging around cant find any good documentation on when in_transit email is triggered.

firstred:
When a tracking number is entered or when a module adds the shipping number and triggers an email. If the email doesn't get triggerd then the module needs to be changed.

getdatakick:
This was resolved by commit d69889e5bc4f2b9ff821bd555b7ebee0798e95af

PR #1751, opened 2024-01-18
Move updateShippingNumber to OrderCarrier class so it can be used by other scripts.

### LINKED ISSUE #1750: in_transit email only sent if working with adminOrdersController
Whole mail::send(in_transit) thing is in controller. It cannot be used outside without duplicating mail send code and managing of email variables.
I think it should be moved to specific OrderCarrier method (commit following soon)

seigieu:
Up until now, any external (via module) update of shipping number would require a module to send in_transit email on its own (and manage variables etc. for each tb version)

Acually had hard time deciding if should use &$errors isntaed of exceptions.
This case has lightErrors (mail not sent) and Heavey errors (invalid number)
Should lightError acually fail triggerEvent? I think not - as reference was updated. So update was a sccess and I don't think it should not triggerEvent, even if email was not sent.

Fixes #1750
Acually may affect #297 and adds shipping_number from #856

### DIFF
--- a/classes/order/OrderCarrier.php
+++ b/classes/order/OrderCarrier.php
@@ -97,15 +97,6 @@ class OrderCarrierCore extends ObjectModel
         ],
     ];
 
-
-    /**
-     * @param $trackingNumber
-     * @param $sendMail
-     * @param $errors
-     * @return bool
-     * @throws PrestaShopException If Customer or Carrier cannot be loaded
-     */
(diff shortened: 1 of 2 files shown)


getdatakick:
Thank you, very nice enhancement.

I've picked this via commit 0955fb897c656b65130d7c48398813a16f08a12c.

Issue #1752, opened 2024-01-20
Accessories returning json instead of list when product is not saved


seigieu:
While adding new product, ajax_products_list.php call have empty excludeIds. It evalutates to false returning ajax instead list for not saved products. For saved products it works ok

PR #1753, opened 2024-01-20
Add parameter to force returnType (ajax|list).

seigieu:
If returnType not defined, then behave like before
Fix issue #1752

### DIFF
--- a/admin-dev/ajax_products_list.php
+++ b/admin-dev/ajax_products_list.php
@@ -58,7 +58,8 @@ if ($pos = strpos($query, ' (ref:')) {
     $query = substr($query, 0, $pos);
 }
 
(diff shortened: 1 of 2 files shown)


getdatakick:
Picked via commit 461afe1bcf500c1e1b9e0d5d14a11b7f7513633c

Provided patch contained extra + symbol on the first line. I had to remove it to avoid 500 error message.

admin-dev/ajax_products_list.php
```diff
@@ -58,7 +58,8 @@
     $query = substr($query, 0, $pos);
 }
 
-$excludeIds = Tools::getValue('excludeIds', false);
+$excludeIds = Tools::getValue('excludeIds', false);+
```
getdatakick, line 61: extra + symbol at the end of line

PR #1768, opened 2024-02-21
Shortcodes in Order Messages (Backoffice)

eschiendorfer:
In our daily business we use order messages quite frequently. But it's always a bit painful to replace placeholders in our messages. That's why I coded a little shortcode replacement with very basic values.

If you merge this, you can use the follwoing shortcodes in order messages:
[customer_firstname], [customer_lastname], [customer_gender], [order_reference], [order_date], [order_delivery_date], [order_total_paid_tax_incl], [order_total_paid_tax_excl], [order_address_delivery], [order_address_invoice]

Example of usage:
```
Dear [customer_gender] [customer_fristname]

We have the feeling that you mistyped your address. Are you sure, its correct like this:
[order_address_delivery]

...
```

Obviously there is much room left to improve it and the customer service in general, but imo is a nice little step in the right direction...

### DIFF
--- a/classes/order/OrderMessage.php
+++ b/classes/order/OrderMessage.php
@@ -97,7 +97,7 @@ class OrderMessageCore extends ObjectModel
             }
 
             if (Validate::isLoadedObject($order) && Validate::isLoadedObject($customer)) {
-                self::replaceShortcodesInOrderMessages($orderMessages, $idLang, $order, $customer);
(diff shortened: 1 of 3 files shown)


getdatakick:
Very nice enhancement, thank you. Picked via commit db85aa271005ab890a8c3a9711a3cca9722d5095

PR #1774, opened 2024-03-13
Allowing order messages on customer service controller

eschiendorfer:
This PR is another attempt to improve customer service without reinventing the whole wheel. It's not logical that you can predefine/select messages only for order. This PR does add this selection also for CustomerService Controller:

image


getdatakick:
Picked via commit 985aa71e1a15492e33a1d558bbc0ffdd6fb8d407

PR #1775, opened 2024-03-13
Small improvement of shortcodes in order messages

eschiendorfer:
I assume that https://github.com/thirtybees/thirtybees/pull/1774 will be merged somehow. To make this really "useful". We also need to remove the order restriction. So that an order message can replace customer shortcodes, even if there is no order linked to the message.


getdatakick:
Implemented slightly different solution in commit 55506b453ecc3555972fa47ef4dea43d559db987

Issue #1744, opened 2023-12-10
Some js files are not being included in windows based systems


pjrp:
I think there is issue in class Media.php with windows based systems.

Apache/2.4.54 (Win64) OpenSSL/1.1.1t PHP/8.0.30

Thirty bees version: 1.6.0
Thirty bees revision: [baa095b098aa369cc0b085b72fb36450bbb06174](https://github.com/thirtybees/thirtybees/tree/baa095b098aa369cc0b085b72fb36450bbb06174)
Build for PHP version: 8.0

My test shop is missing inclusion of some js files causing issue in the frontend and admin themes. 

For example:

- files in the js/autoload theme folder
- controller.js from the core updater v1.6.6.

Done some testing in function getLocalMediaFilePath($uri), i got a sample of the offending URIs:

"C:\laragon\www\shop/themes/niara/js/autoload/10-bootstrap.min.js"
"C:\laragon\www\shop/modules/coreupdater/views/js/controller.js"
...

I was unable to find how _PS_ROOT_DIR_ is added to some js files and not others.

Could we fix this by removing _PS_ROOT_DIR_  before calling getLocalMediaFilePath() ?

$mediaUri = str_replace(_PS_ROOT_DIR_, "",  $mediaUri);

getdatakick:
While thirty bees is officially not supported on windows, I believe this should be fixed anyway.

Anyone with windows box, please try to fix and provide PR. I don't have access to windows machine myself :)

deweblooper:
This can be a solution,
I´ve made it with comparing results of function on Windows and Linux server, it works without issues,
can be used as override, or if you decide to pull it to official TB, I will be happy.

```
class Media extends MediaCore
{

	/**
	 * for different server environments
	 */
    public static function getLocalMediaFilePath($uri)
    {
        if (! $uri) {
            return false;
        }

        $uri = (string)$uri;

        // if file exists locally, include its modification timestamp into uri as a version parameter
        $parsed = parse_url($uri);
        if (!array_key_exists('host', $parsed) && isset($parsed['path'])) {
            $path = $parsed['path'];
            $rootDir = rtrim(str_replace('\\', '/', _PS_ROOT_DIR_), '/');
            $filePath = $rootDir . '' . $path;	// deleted slash separator as unnecessary in both environments
            if (file_exists($filePath) && is_file($filePath)) {
                return $filePath;
            }
[...]

PR #1790, opened 2024-05-06
Media Class fix for Linux and Windows compatibility

deweblooper:
There is an issue with function getLocalMediaFilePath() caused by directory separator used in Windows different from one in Linux environments. This causes ignoring and skipping important files like autoload js and css in theme on Windows based servers.
Issue is documented in https://github.com/thirtybees/thirtybees/issues/1744 .
This fix detects environment and assemble propper url

### DIFF
--- a/classes/Media.php
+++ b/classes/Media.php
@@ -1054,6 +1054,7 @@ class MediaCore
 
     /**
      * Returns full path to local file for $uri, or false if file does not exists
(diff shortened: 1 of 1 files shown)


getdatakick:
Thank you, picked via commit bf2fe6fcbed5b2fc29504ca4a0ac044705bcc1bf

Issue #642, opened 2018-10-29
Make Shopping Carts sortable and searchable


Traumflug:
See https://forum.thirtybees.com/topic/2196/sort-cart-controller-by-total-amount

PR #1804, opened 2024-05-25
AdminCartsController - sortable carts, delete empty carts and delete …

therampagerado:
…carts older than %d days, plus attempted bugfix for deleted products

**Changes:**

**'Total' column is now sortable** - stolen from https://forum.thirtybees.com/topic/2196-2196/sort-cart-controller-by-total-amount/#comment-22086 and related to https://github.com/thirtybees/thirtybees/issues/642

**Two new buttons** - 'Delete empty carts' and ' Delete carts older than %d days' related to https://github.com/thirtybees/thirtybees/issues/776

Attempt on fixing an issue where carts show 0.00 if all products are deleted (old carts) even if order resulted from this cart (green background). I think I managed to fix it in the list but I was unable make it work in cart summary.

Please, review the code when you have time. Thanks!

### DIFF
--- a/controllers/admin/AdminCartsController.php
+++ b/controllers/admin/AdminCartsController.php
@@ -145,14 +145,34 @@ class AdminCartsControllerCore extends AdminController
      * @throws PrestaShopException
      */
     public static function getOrderTotalUsingTaxCalculationMethod($idCart)
-    {
-        $context = Context::getContext();
-        $context->cart = new Cart($idCart);
-        $context->currency = new Currency((int) $context->cart->id_currency);
-        $context->customer = new Customer((int) $context->cart->id_customer);
+	{
+		$context = Context::getContext();
+		$context->cart = new Cart($idCart);
+		$context->currency = new Currency((int) $context->cart->id_currency);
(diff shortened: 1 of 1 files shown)


getdatakick:
Re - sortable totals:

I will not merge this commit (21921b9977b5812c7d779ad9c318c524501187f7), as it doesn't work as expected. The total amounts are not calculated in list SQL query. Totals are instead calculated in callback mechanism when rendering list results. 

This change actually sorts rows by id_cart instead.

getdatakick:
The last commit related to displaying price of cart with missing products does not work properly. There are few things that should be addressed:

1) method `Cart::getTotalCart` returns string with formatted price. It can be **$0.18** or **1,000,029.23€** or anything else. Comparison `$cartTotal > 0` wont work correctly in all cases
2) Fallback to order total should respect tax settings - display prices with or without tax
3) this solution works only when all products in the cart are no longer in db. If only a subset of products are missing, the total price will be incorrect

I'm not sure how I feel about displaying order / historical prices . The semantics of cart list/detail is to display current value of products in cart. From this point of view, it does makes sense to display zero if cart contains products that no longer exists.

PR #1811, opened 2024-06-05
ShopMaintenance - delete old lines in /error_log and /admin/error_log

therampagerado:
Reused getLogsRetentionPeriod constant as it basically applies to those logs too.

Delete old lines and leaves those after the cutoff date.

Line by line approach for old uncleaned files that could be MB in size - should save RAM on first run.

This method appears to work with errors that don't start with a timestamp (Mail queue module) as it checks the next line for retention and removes anything above it. So even if the file is manually edited this approach should handle it.

Restores last modified date for easier debugging when looking purely on the date.

### DIFF
--- a/classes/ShopMaintenance.php
+++ b/classes/ShopMaintenance.php
@@ -46,6 +46,7 @@ class ShopMaintenanceCore
             static::cleanAdminControllerMessages();
             static::cleanOldLogFiles();
             static::cleanOldThemeCacheFiles();
+            static::deleteOldErrorLogEntries();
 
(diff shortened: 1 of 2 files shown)


getdatakick:
I'm sorry, but I don't think this should be part of the core.

These `error_log` files are NOT created by thirty bees application. They are created directly by php server (apache, php-fpm, or whatnot), and as such they should not be modified by php app itself.

Location and format of php error log file can be tweaked in php.ini configuration. Log rotation could and should be configured on system level as well. 

If your hosting puts error_log files inside your root directory, ask them to reconfigure it. This is actually quite serious security risk, as raw php errors can contain sensitive informations.

PR #1813, opened 2024-06-11
Add button to close threads older than 30 days

therampagerado:
By default any order message generates a thread. This could be turned off but the main issue persist - there is no way to automatically close threads older than NN days (a normal practice in all customer service systems).

This fix introduces a simple method to close all 'open' threads in AdminCustomerThreadsController.


getdatakick:
Picked via commit 7fc1467218a61d83198ded51d1a2d77d4804bb3e.


Thank you

PR #1894, opened 2024-09-27
Update jquery.bxslider.js

Zengraph:
Updating to last 4.2 version : v4.2.1d
Fix some bugs, specially when product variant selected on product page and slider got stucked.

### DIFF
--- a/js/jquery/plugins/bxslider/jquery.bxslider.js
+++ b/js/jquery/plugins/bxslider/jquery.bxslider.js
@@ -1,10 +1,1663 @@
 /**
- * BxSlider v4.1.2 - Fully loaded, responsive content slider
- * http://bxslider.com
- *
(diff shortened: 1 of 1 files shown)


therampagerado:
@Zengraph isn't this the latest file version?

https://github.com/stevenwanderski/bxslider-4/blob/master/src/js/jquery.bxslider.js

getdatakick:
Picked via 2e65335cdbc1145be0d945ab4bf43bc64cd01fe3

Zengraph:
Hello,

"Do we know how if there are any breaking changes between 4.1.2 and this 4.2.1d ?" => it seems that it's just bugfixes after using it several days, but I didn't check all the code of this file.

"Is there a reason why the latest version 4.2.7 is not used instead of 4.2.1d?" => nop, it was just the closest version from the one used that i picked up, hoping that there will be no breaking changes or the less possible. But it might be better to use the 4.2.7 it that works even better.

Best Regards

PR #1910, opened 2024-10-20
1907: import only product combinations

elitecolors:
instead of truncate all products combinations we just delete combinations for imported product

### DIFF
--- a/controllers/admin/AdminImportController.php
+++ b/controllers/admin/AdminImportController.php
@@ -1247,11 +1247,7 @@ class AdminImportControllerCore extends AdminController
         // Check if the CSV file exist
         if (Tools::getValue('filename')) {
             $entityType = $this->getSelectedEntity();
-            $shopIsFeatureActive = Shop::isFeatureActive();
-            // If i am a superadmin, i can truncate table (ONLY IF OFFSET == 0 or false and NOT FOR VALIDATION MODE!)
-            if (!$offset && !$moreStep && !$validateOnly && (($shopIsFeatureActive && $this->context->employee->isSuperAdmin()) || !$shopIsFeatureActive) && Tools::getValue('truncate')) {
-                $this->truncateTables($entityType);
-            }
+
             $doneCount = 0;
             // Sometime, import will use registers to memorize data across all elements to import (for trees, or else).
             // Since import is splitted in multiple ajax calls, we must keep these data across all steps of the full import.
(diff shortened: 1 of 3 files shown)


controllers/admin/AdminImportController.php
```diff
@@ -3629,6 +3625,13 @@ public function attributeImport($offset = false, $limit = false, &$crossStepsVar
         $regenerate = Tools::getValue('regenerate');
         $shopIsFeatureActive = Shop::isFeatureActive();
 
+        $processedProducts = [];
+        $truncateOnlyProductCombinations = false;
+
+        if (!$offset  && !$validateOnly && (($shopIsFeatureActive && $this->context->employee->isSuperAdmin()) || !$shopIsFeatureActive) && Tools::getValue('truncate')) {
+            $truncateOnlyProductCombinations = true;
```
getdatakick, line 3632: This commit changes semantics of `Delete all combinations before import` functionality.

We should keep that functionality in place, because some merchants actually might want to truncate all combinations before import new one. 

There should be separate option for this functionality in import settings

image

controllers/admin/AdminImportController.php
```diff
@@ -3643,6 +3646,10 @@ public function attributeImport($offset = false, $limit = false, &$crossStepsVar
             $info = array_map('trim', $info);
 
             try {
+                if ($truncateOnlyProductCombinations) {
+                    $this->deleteCombinationsByProductId($info['id_product'], $processedProducts);
+                }
```
getdatakick, line 3651: Unfortunately this won't work. The process of big files is split into multiple separate request, each of them processing 5 rows at a time. Your `$processedProducts` array will be empty at the start of each request.

You need to, somehow, make this $processedProducts persistent. Maybe use `$crossStepsVariables`

getdatakick (changes requested):
Thank you for the PR. Before I merge it, I need you to fix the two problems:

- do not remove original 'truncate all' functionality, as it is used by some merchants
- fix the import for large files, which are split into multiple ajax request

controllers/admin/AdminImportController.php
elitecolors: new field added probably we will need translation etc

getdatakick:
Picked via commit 4ef822ecbac35d71126f77567fc81b9eb5289c2b.

I've slightly modified the code -- to delete all combinations, we use existing `Product::deleteProductAttributes` method. This method performs some post-processing tasks (trigger hooks, regenerate specific prices, etc)

PR #1917, opened 2024-10-31
Fix 500 error on module controller wrong case call

metacreo:
Fix 500 error output when call module controller in wrong case.

Try:
example.com/en/module/mailalerts/**a**ccount
example.com/en/module/mailalerts/**A**ccount

This bug allow to init 500 error on server just from address bar.
---
edit: Also you can rename account.php to Account.php then try call again and see behavior.


getdatakick:
Thank you.

Issue #1920, opened 2024-11-07
Remove ISO code from the URLs for the first (default) language


Yabbertb:
If you launch a store with only one language, then all the URLs in the store do not have the language suffix after the domain.
If the store, after some time of operation, decides to add a new language, all the URLs of the default language are spoiled and the ISO suffix of the language is added to them. With a large number of products in the store, this requires the creation of thousands of 301 redirects for the old URLs, and even so, the positions of these pages record large drops in the SERP results of search engines.
There are modules for prestashop that remove the ISO code for the default language, but I think such a serious bug should be fixed at the core level.

getdatakick:
I don't see this as a bug. This behaviour existed forever.

I agree that we could implement enhancement - settings option with three options for language iso code in url

1) current behaviour -- include iso code for multi-language shops, hide iso code for single language shops
2) always include show iso code, even for single-language shops
3) include iso code only for non-primary language

#### Workaround: 
if you need to add new language, you can always opt for multistore. Add a new store that shared everything with your primary store, and associate this store with single language only. Then use subdomains for every store:

- www.domain.com
- it.domain.com
- cz.domain.com

Yabbertb:
> I don't see this as a bug. This behaviour existed forever.

There have always been and still are thousands of serious bugs in prestashop.
Destroying all the URLs in the store, just because of adding a new language, is certainly a serious bug.

If we decide to use national domains like mydomain.uk, mydomain.de, mydomain.fr then multistore solution is a good solution.
But running subdomains like en.mydomain.com, de.mydomain.com, fr.mydomain.com is not the best solution for SEO reasons.

getdatakick:
> There have always been and still are thousands of serious bugs in prestashop. Destroying all the URLs in the store, just because of adding a new language, is certainly a serious bug.

This is by design, hence not a bug :)

> If we decide to use national domains like mydomain.uk, mydomain.de, mydomain.fr then multistore solution is a good solution. But running subdomains like en.mydomain.com, de.mydomain.com, fr.mydomain.com is not the best solution for SEO reasons.


In multistore you can assign any domain you wish. You can use national domains, or subdomains, its up to you.

getdatakick:
@Yabbertb

This enhancement has been implemented in branch issue-1920. You can test it (it's possible to update to the branch using coreupdater, or you can install it manually). 

Since this is a big change, I'd like for more people to test it before I merge it to main

Yabbertb:
[...]

PR #1932, opened 2024-12-03
Remove ISO from default language URL when multilang.

therampagerado:
Related to: https://github.com/thirtybees/thirtybees/issues/1920

Known bug:
- when switching from default to non-default language - it is working
- when switching back - the shop is stuck on the non-default language.

### DIFF
--- a/classes/Link.php
+++ b/classes/Link.php
@@ -293,14 +293,21 @@ class LinkCore
             $context = Context::getContext();
         }
 
-        if ((!$this->allow && in_array($idShop, [$context->shop->id, null])) || !Language::isMultiLanguageActivated($idShop) || !(int) Configuration::get('PS_REWRITING_SETTINGS', null, null, $idShop)) {
-            return '';
-        }
-
         if (!$idLang) {
             $idLang = $context->language->id;
         }
 
+        // Check conditions for returning an empty string in the URL
+        if ((!$this->allow && in_array($idShop, [$context->shop->id, null])) 
+            || !Language::isMultiLanguageActivated($idShop) 
+            || !(int) Configuration::get('PS_REWRITING_SETTINGS', null, null, $idShop)
+            || ($idLang == Configuration::get('PS_LANG_DEFAULT') 
+                && Configuration::get('TB_REMOVE_DEFAULT_LANGUAGE_ISO_URL', null, null, $idShop))
+        ) {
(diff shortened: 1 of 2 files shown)


getdatakick:
Hi,

thank you for the PR. I've used it as a base code, and extended it a little bit to include additional options.
I've pushed in in branch issue-1920. Please help me test the solution -- it is possible to update to that branch using core updater.

I've also added an option to change language friendly url -- it no longer needs to be language iso code. Probably nobody will use it, but I think it's a nice touch.

therampagerado:
Thank you,

Will test it today.

therampagerado:
@getdatakick 

I did a short test and found 2 bugs that are probably related - you fixed the same behavior for my module doing the same last month.

1. When on Authentication page and non-default language the form is loaded in the default lang. When you get out of this page the site language is switched to default.
2. On a category page when you select Layered Nav filter it loads the default translation, URLs, everything. When you get out of this page the site language is switched to default.

You did some cookies magic to fix those issues.

A demo: 

1. https://test2.rampagesport.eu/login?back=my-account - switch to BG - you see 'Създай акаунт' (Create an account), when you click on the button it loads the default lang version (EN in this case).
2. https://test2.rampagesport.eu/gifts - click on "Производител" (Manufacturer) filter and look it loads the filter in EN, not in BG. (for the layered nav only this is translated so the rest of the titles will be in EN always).


I'll continue with more manual tests that I can think of, but overall I think it works perfectly.

therampagerado:
3. Breadcrumbs home link is always default language in non-default language.


This bug is present on my production too, I have not noticed it until now.

getdatakick:
Thanks for testing.

Problems 1 and 2 were fixed in core. See description in commit ada42cf5e6aab6d3f1fbcff56ce6c38b7ec356b4

Problem 3 -- this is a theme issue. Theme simply redirects user to `/` and, and expect thirty bees to pick up current language from cookie and redirect to language variant. This no longer works, because this url `/` is now valid even in multistore env. There is no fix for core -- the theme needs to be fixed.

See commit in niara: https://github.com/thirtybees/niara/commit/ceecca83a97b2b5f12340bc8c6b33ed7e0211eca

getdatakick:
Now this is part of bleeding edge. Thanks again for original code, and for testing

Issue #1936, opened 2024-12-10
AdminDuplicateUrls: checking duplicates from Meta class


Yabbertb:
If the store administrator makes a mistake and adds duplicate URLs in the Meta class, we have a malfunctioning store. If we have several languages in the store, then visually determining the error of duplicate URLs is not even possible.
Also, it would be very useful to check duplicate URLs from the Meta class in AdminDuplicateUrls.

therampagerado:
Will something like this suffice? 
![image](https://github.com/user-attachments/assets/05370569-9651-4779-ad3f-6f9ea184ecd9)

It detects the pages that don't need additional parameters (like Shops, Contact us, Best sellers) and matches them against all other cases.

PR #1937, opened 2024-12-11
Support for Meta pages duplicate SEO URL detection.

therampagerado:
Please, review the URL building for Meta case as I feel there's a better way.

Related to https://github.com/thirtybees/thirtybees/issues/1936


getdatakick:
Thank you!

Picked via commit e972a6f6e67424a36da912ab67cdb38b58825e12.

I've slightly changed the logic to resolve page url. It now uses `Link::getPage` method, but only for routes that do not contain required parameters.

PR #1938, opened 2024-12-12
TinyMCE codemirror plugin

Yabbertb:
Standard "code" plugin for TinyMCE only makes editing HTML code more difficult. I suggest replacing it with the "codemirror" plugin.


getdatakick:
Nice, thank you.

PR #1946, opened 2025-01-12
Allow to override ImageManager's resize method.

moaai:
This update allows you to override the ImageManager resizing method.

Reason
I have a plenty of images of varying sizes and aspect ratios and I found that the custom zoom/crop functionality is the best option for my store. I couldn't achieve the same behavior using the default resize method, hence this update.


getdatakick:
Nice catch, thank you. We should always use `static` binding to support overriding

Issue #1952, opened 2025-01-22
Multistore: robots.txt file is only one and static


therampagerado:
If we use static robots.txt file it is one for every shop when using multistore.

We can't add different sitemaps to it and we can't have different rules per shop.

A possible solution is to generate it dynamically:

1. htaccess:

`RewriteRule ^robots\.txt$ robots.php [L]`

2. robots table to store the contents per shop:
`CREATE TABLE IF NOT EXISTS `tb_robots` (
    `id_shop` INT(10) UNSIGNED NOT NULL,
    `robots_content` TEXT NOT NULL,
    PRIMARY KEY (`id_shop`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;`

3. robots.php:
```
<?php
require_once dirname(__FILE__).'/config/config.inc.php';

// Determine current shop context
$shop = Context::getContext()->shop;
$shop_id = $shop->id;

// Fetch robots.txt content from database
$robots_content = Db::getInstance()->getValue(
    'SELECT robots_content FROM tb_robots WHERE id_shop = ' . (int)$shop_id
);

if ($robots_content) {
    header('Content-Type: text/plain');
    echo $robots_content;
} else {
    // Fallback: Return a default robots.txt
    echo "User-agent: *\nDisallow: /";
}

```

PR #1953, opened 2025-01-22
Dynamic robots.txt generation

therampagerado:
Generate dynamically the robots.txt file.

Current approach does not work with Multistore, this approach offers possibility to have per-shop rules and also dynamically pulls the correct sitemap to be included.

All big CMSes use dynamic robots.txt

Related to https://github.com/thirtybees/thirtybees/issues/1952

### DIFF
--- a/classes/Tools.php
+++ b/classes/Tools.php
@@ -3227,7 +3227,8 @@ class ToolsCore
         }
 
         fwrite($write_fd, "RewriteEngine on\n");
-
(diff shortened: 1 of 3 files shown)


getdatakick:
I'm not opposed to something like this. However, the implementation needs to be rewritten. As it is now, it depends on apache rewrite functionality. A lot of thirty bees store do not run on apache, but for example on nginx+phpfpm.

In order to have this, it should be implemented as standard FrontController. Page with (immutable) route `/robots.txt` should be created for it in `Preferences > SEO & URLs`

PR #1955, opened 2025-01-25
Dynamic robots.txt - attempt 2

therampagerado:
A second attempt, hopefully better.

1. Migration starts by accessing index.php?controller=robots
- this creates the needed records for tables meta, meta_lang, robots and a record in Configuration for completed migration and setup.
2. The migration will look for robots.txt, copy it's contents in robots table and correct the sitemap url per shop (currently it's same for all shops). At the end a record is set that the migration is done and original robots.txt file is renamed robots.txt.old
3. We're using dispatcher to forward all calls to robots.txt to RobotsController which serves values from the table or if not present (migration is uncomplete because of missing robots.txt altogether) it serves conservative code and logs an error in BO so merchant hopefully sees it and generate the table entries by using the button.

I have the following bugs.
1. Migration should be started manually by accessing index.php?controller=robots. If this is not done robots.txt will continue to be served with precedence. I am unable to think how this can be avoided without using redirects.
2. Robots.txt contents in SEO & URLs acts funny and displays old/cached info, not the actual table entry upon refresh. I'm unable to find the bug.
3. I don't know how to add robots table to the golden files for new installs.
4. I don't know how to log to CollectLogs module too.

Potential issues:
1. SEO Optimization modules for PS could look for robots.txt in order to edit it. No such modules as far as I know for thirtybees for now.

### DIFF
--- a/classes/Dispatcher.php
+++ b/classes/Dispatcher.php
@@ -753,6 +753,17 @@ class DispatcherCore
         switch ($this->front_controller) {
             // Dispatch front office controller
             case static::FC_FRONT:
+            
+                 // Handle robots.txt without relying on URL rewriting
+                if (strtolower($_SERVER['REQUEST_URI']) === '/robots.txt') {
+                    require_once _PS_ROOT_DIR_ . '/controllers/front/RobotsController.php';
+
+                    // Instantiate RobotsController directly
+                    $controller = new RobotsController();
+                    $controller->run();
+                    exit; // Stop further processing
+                }
+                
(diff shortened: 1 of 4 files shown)


getdatakick:
How about we created a module instead of core change? Not everyone needs this, in fact for most people static robot.txt file is enough. People that need dynamic robots.txt file could simply install the module.

Having this as a module would solve all your issues -- initial setup would be done during module installation, and existing robots.txt file could be deleted/converted to db entries.

There is no need to modify Dispatcher.php. The only disadvantage I can see is the need to override `generateRobotsFile`, and make it no-op.

therampagerado:
OK, I'll send you the module.

PR #1957, opened 2025-01-29
FIx combinations import with large files

therampagerado:
When importing large combination csv files I get this error:

image

With those changes I am able to import those files.


getdatakick:
After investigation it was determined that browser sent `crossStepsVars` post parameter, but server failed to parse it. This PR works around that by hiding this fact, but that's not a proper way to deal with it.

PR #1962, opened 2025-02-11
getCrossSells consider product_store active

akuehnis:
Modification of OrderDetail::getCrossSells


getdatakick (commented):
`active` is a multi-store property, so the condition should use `PREFIX_product_shop` table

getdatakick (changes requested):
Could you please change the PR to use product_shop table instead of product

akuehnis:
Yes, of course. Sorry.

getdatakick:
Picked via commit 1bce251edaad9f004aeab496dd4974ab5dcb787b

PR #1977, opened 2025-03-17
Automatic DB backup.

therampagerado:
Introduce automatic daily DB backup mechanism in ShopMaintanance plus kill switch and retention period setting in AdminBackupController.


getdatakick:
Thank you.

In the future, I would like to convert all tasks in ShopMaintanance to full-fledged WorkQueue tasks. That way, we can schedule them, and execute them from different context (for example from CLI). 

But for the time being, having this in ShopMaintanance is better than nothing.

PR #1988, opened 2025-03-30
AdminOrders: display pack items

Yabbertb:


getdatakick:
Hi,

it's a good idea, but the implementation is not correct.

You are displaying current pack items, but you should be displaying items that were part of the pack at the time of a sale. 

We keep information about that in table `tb_order_detail_pack`

controllers/admin/AdminOrdersController.php
```diff
@@ -1955,6 +1955,16 @@ public function renderView()
                 $product['warehouse_name'] = '--';
                 $product['warehouse_location'] = false;
             }
+            if (Pack::isPack($product['product_id'])) {
+                $productPackItems = Pack::getItems($product['product_id'], $this->context->language->id);
```
getdatakick, line 1959: you should fetch pack content from `tb_order_detail_pack`

getdatakick (changes requested):

Yabbertb:
Hello.

The first concept was indeed wrong, because if the store changes the components of the package after the order, the order will display incorrect data.
I have extended the OrderDetailPack class with data retrieval methods, because this class is completely empty.
We can use these methods, for example, to generate a PDF document "Delivery slip", because now the warehouse does not know what to pack into packages.

classes/order/OrderDetailPack.php
```diff
@@ -62,4 +62,171 @@ class OrderDetailPackCore extends ObjectModel
...
+     * Get a order detail Id
+     *
+     * @param int $idOrder
+     * @param int $idProduct
+     * @return int
+     * @throws PrestaShopDatabaseException
+     * @throws PrestaShopException
+     */
+    public static function getOrderDetailId($idOrder, $idProduct)
+    {
```
getdatakick, line 212: This is probably not correct. There can be multiple records in `order_detail` for the same product ID. For example when product customization is used

getdatakick (changes requested):
This sill not work with order that have product customization, or if the pack has combinations (new upcoming feature).

Yabbertb:
Hi,
Yes, the _customized_data.tpl template modification was missing (I've already added it).
If a customer creates an order and adds several of the same packages with customization, each of these packages has the same components (the customer cannot change the components of the package), so it doesn't matter that a product with the same ID appears multiple records in order_detail.

I haven't installed and tested this new "pack with combinations" functionality yet. I'll find some time and take a look at it.

getdatakick:
> If a customer creates an order and adds several of the same packages with customization, each of these packages has the same components (the customer cannot change the components of the package), so it doesn't matter that a product with the same ID appears multiple records in order_detail.

This is an assumption that might be currently valid, but can very easily be changed in the future. For example, that `pack with combinations` functionality actually changes the pack content depending on selected pack attribute.

I say don't introduce any assumptions into the code. Instead of depending on product id like this: 

```php
if (OrderDetailPack::isPack($order->id, $product['product_id'])) {
    $productPackItems = OrderDetailPack::getItems($order->id, $product['product_id'], $this->context->language-
}
```

we can simply work with OrderDetail ID, and retrieve pack content for every line:

```php
if (OrderDetailPack::isPack($order->id, (int)$product['id_order_detail'])) {
    $productPackItems = OrderDetailPack::getItems($order->id, (int)$product['id_order_detail'], $this->context->language-
}
```

getdatakick:
> I just don't know if it will be compatible with this new "pack with combinations" functionality you plan to add.

It will be

Thank you. This is great addition, I'll merge this

getdatakick:
Picked via squash commit 543dd8d1397aea0df3a801e99df7aca655491ec8

Yabbertb:
Thanks.

Now I will add printing of package items on PDF "Delivery slip".
We can also think about displaying package items on the front-office in the customer order history :)

PR #1994, opened 2025-04-10
Db backup fix

therampagerado:
Previously a concurrent execution was possible when page refreshes during active backup or if multiple employees are logged at the same time.

### DIFF
--- a/classes/ShopMaintenance.php
+++ b/classes/ShopMaintenance.php
@@ -183,11 +183,35 @@ class ShopMaintenanceCore
      */
     public static function autoDbBackup()
     {
-        if (Configuration::get('TB_DB_AUTO_BACKUP')) {
+        if (Configuration::get('TB_BACKUP_RUNNING')) {
+            return;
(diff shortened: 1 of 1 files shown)


getdatakick:
This will not work properly. The configuration table is loaded once at the beginning of request, so the check `Configuration::get('TB_BACKUP_RUNNING')` is not atomic.

You should rewrite this to use database lock. You can take an inspiration here: 

https://github.com/thirtybees/thirtybees/blob/79049ca4ba75fdc556e905b4ca141370610c9660/classes/workqueue/Scheduler.php#L140-L146

Also, I'm not a big fan of catching exceptions and logging them to PrestaShopLogger. That just hides them -- nobody ever looks into Logs. If there is an exception, let it bubble out - that way, it will be catched by error handler and processed (passed to in collectlogs module, logged in /log/ directory, or any other error handling tool that might be installed)

therampagerado:
Hope this works better.

PR #1996, opened 2025-04-16
New Product Core method to reliably get product stocks ready for sale.

gregzawadzki:
In many cases quantity can be 0 even if product has stocks. For example product has one attribute of 20 and other one of -20. Summing it to 0 which is false.

This method allows to more reliably determine if this product has any quanity for sale.

### DIFF
--- a/classes/Product.php
+++ b/classes/Product.php
@@ -4130,7 +4130,7 @@ class ProductCore extends ObjectModel implements InitializationCallback
         $result = $conn->execute('
             UPDATE `'._DB_PREFIX_.'category_product`
             SET `position`= `position` '.($way ? '-1' : '+1').'
-            WHERE `position` '.($way ? '>': '<').$currentPosition.' 
+            WHERE `position` '.($way ? '>': '<').$currentPosition.'
               AND `position` '.($way ? '<=' : '>=').$newPosition.'
               AND `id_category` ='.$categoryId
         );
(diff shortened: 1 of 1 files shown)


metacreo (approved):
This is great! 
I use a similar method in many places. It significantly reduces the amount of code and allows you to avoid manually calling another class. I think this is a useful feature, although it requires a little memory expansion.

gregzawadzki:
Kontakt


    
        Dziękujemy za Twoją wiadomość!
        Ten adres e-mail istnieje historycznie i nie jest regularnie sprawdzany. Odpowiedź na Twoją wiadomość może zająć od kilku dni do kilkunastu dni.
        Aby uzyskać szybszą odpowiedź, prosimy o skorzystanie z formularza kontaktowego na naszej stronie internetowej:
        
            Przejdź do formularza kontaktowego
        
        Dziękujemy za zrozumienie!

gregzawadzki:
Glad you like it. Hope it will replace template checking against `$product->quantity > 0` in future.

classes/Product.php
```diff
@@ -8435,4 +8435,51 @@ public function getDepth(int $combinationId = 0): float
...
+     * @param $ignoreNegativeStocks bool If should ignore negative stocks
+     * @param $fresh bool If true, then cache is refreshed
+     * @param $idShop int|null ID shop for which to take stocks from
+     * @return int
+     * @since thirty bees 1.7.0
+     *
+     */
+    public function getAvailableQuantity(bool $ignoreNegativeStocks = true, bool $fresh = false, int $idShop = null): int {
+        static $cache = [];
+        $key = ($ignoreNegativeStocks ? 'p' : 'n') . '_' . $idShop;
```
getdatakick, line 8454: cache key must include product id to avoid collision

getdatakick:
It's definitely useful method. I found a small bug in your PR related to caching. Please fix it, and then I'll merge your code.

Thank you

classes/Product.php
gregzawadzki: Nvm. See my mistake.

getdatakick:
Picked via commit c01d999941538b277a940587aa927439bdc30eeb

PR #2004, opened 2025-05-08
Update Controller.php

metacreo:
This sortable script too old not working and only cause error in dev console when enabled debug and profiling.
sortable


getdatakick:
Picked by commit 5875673cba51c441c67d2d9f56158a07be021110. Thank you

PR #2007, opened 2025-05-15
Upgrade Cart.php > getProducts improvement

metacreo:
The getProducts function returns a list of products with parameters. Some of these parameters are not used anywhere at all and are actually useless.

This modification changes several of these parameters so that a promising practical opportunity to use them in templates appears. 

For example, in the Ajax cart you can now display discounts (-15% or -26$) and show messages to the users like "add to cart N qty to cart to apply discount.".

Without additional queries from template or module.

### DIFF
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -643,8 +643,16 @@ class CartCore extends ObjectModel
                 }
             }
 
-            $row['reduction_applies'] = ($specificPriceOutput && (float) $specificPriceOutput['reduction']);
-            $row['quantity_discount_applies'] = ($specificPriceOutput && $quantity >= (int) $specificPriceOutput['from_quantity']);
+            $row['reduction_applies'] = null;
+            if ($specificPriceOutput && $specificPriceOutput['reduction'] && $specificPriceOutput['reduction_type']) {
+                if ($specificPriceOutput['reduction_type'] == 'amount') {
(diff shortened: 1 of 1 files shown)


getdatakick:
We can't accept this change, as it changes semantics for those array keys. There can be third party (or even native) modules that depends on those values.

For example, your PR changes `quantity_discount_applies` from `boolean` to `int`. Semantics of that value used to be `exists quantity discounts for requested $quantity?`, and newly it is `from quantity of discount`. This can cause backwards compatibility issues with some modules or themes.

If you need more information, please add new array keys with them, don't change semantics of existing one.

getdatakick:
As an example, see this code in third party module:

https://github.com/PaysonBySvea/PaysonCheckout-Prestashop-8/blob/874f7fd8556ce0b7402d0db4b5afc7cee368ee57/modules/ps_paysoncheckout2/ps_paysoncheckout2.php#L1416

That module code would not work properly with your PR.

metacreo:
Thanks for explain.
I didn't take into account third-party modules. I only checked the native ones.
I just wanted to add more functionality. If I leave the existing keys and add a few new ones, is it possible to accept? Just a few additional parameters will significantly expand the functionality of the cart without any performance costs.

getdatakick:
> If I leave the existing keys and add a few new ones, is it possible to accept? 

Yes, of course. Adding additional keys are perfectly fine, and I'm sure there's next to nothing performance cost.

PR #2018, opened 2025-05-21
Improve trigger_error: Attempt to update unsaved object

eschiendorfer:
I like this error message a lot. It helps to find logical bugs/flaws in you code. But I guess it would be more helpful if we would trigger it depending on the class name.


eschiendorfer:
Sorry something went wrong. I will do it again.

PR #2029, opened 2025-06-14
Correct display of “description” block

Yabbertb:


metacreo:
this is already fixed by: Fix broken view of meta description in BO Category editing #1983

PR #2038, opened 2025-07-22
Update form.tpl in backoffice when creating order manualy

metacreo:
Fix image oversize overscreen in BO.
In BO, creating order manualy product image not limited by size.
this images row(td) not have any limits. If your image is large, image shows on all screen.


getdatakick:
Merged, thank you

PR #2057, opened 2025-08-20
Fix dynamic image generation always being square

gregzawadzki:


getdatakick:
Lol, that's a nice bug :)

Thank you for the fix

PR #2059, opened 2025-08-20
AdminCustomersController - fix Default customer group list

therampagerado:
Default customer group should show only currently selected groups to choose from. Now it shows all groups and if we choose a one not associated with the account and save there is an error.

- optimized the logic of the dropdown list
- introduced ajax call to dynamically update the list based on the marked checkboxes of Group access list (no need to save and reenter the account profile to choose the new default)
- notification popup for error/success


getdatakick:
Awesome, thank you.

PR #2081, opened 2025-12-12
Add configuration option for rendering PDFs inline instead of as attachment

mwrouse:
Add a configuration option for allowing PDFs to render as inline instead of attachment. 

We frequently print delivery slips, it is annoying to download each one to print. This PR adds a configuration option under Preferences > General for rendering PDFs as inline instead of as attachment. 

I am unaware if I missed any locations.


getdatakick:
Thank you, merged via commit 00b9f48f43e73fdeb2b58ada72afff57425e7610

PR #2085, opened 2025-12-18
New hook for rendering buttons in the admin orders page

mwrouse:
New hook for adding custom buttons to the admin orders page!
image

There is already a hook for adding to the tabs, I believe there should also be a hook for modules to register buttons on this page. 

As you can see I've added a couple custom buttons myself, most notably a direct link to the paypal/strip order, as well as a button to launch a window for generating a shipping label.


getdatakick:
Merged, thank you

PR #2101, opened 2026-03-01
Allow back-office order creation from guest carts

### LINKED ISSUE #1100: Adding abandoned cart
As shop maintainer you can create an order for a customer. You can also take a cart he abandoned and convert it to an order. However, you can only do that when he was already logged in when he created that cart. You cannot take a random "guest" cart without an owner and convert that for a customer. This means that if the customer didn't log in you still need to retype the order.

It would be nice if you could also add such carts.

therampagerado:
Closes: https://github.com/thirtybees/thirtybees/issues/1100

Motivation

Back-office users must be able to start the order creation flow from any cart (including visitor/guest carts) and then select or create a customer on the order page instead of retyping the cart.

The previous logic prevented opening the order-create form for carts without id_customer and hid the "Create an order from this cart." action for guest carts.

Description

Show the "Create an order from this cart." button for all carts by removing the template condition in admin-dev/themes/default/template/controllers/carts/helpers/view/view.tpl so guest/visitor carts expose the action.

Remove the early renderForm() validation in controllers/admin/AdminOrdersController.php that rejected carts with no id_customer, allowing the order form to load and let the user pick or create a customer there.

Add a submit-time guard in AdminOrdersController::postProcess (the submitAddOrder branch) to require a selected customer before finalizing the order, returning the error You must select a customer before creating the order. if none is present.

### DIFF
--- a/admin-dev/themes/default/template/controllers/carts/helpers/view/view.tpl
+++ b/admin-dev/themes/default/template/controllers/carts/helpers/view/view.tpl
@@ -71,9 +71,7 @@
 				{l s='Made on:'} {dateFormat date=$order->date_add}
 			{else}
 				<h2>{l s='No order was created from this cart.'}</h2>
-				{if $customer->id}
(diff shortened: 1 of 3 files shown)


getdatakick:
Merged. Thank you.

PR #2109, opened 2026-04-01
Fix namespace hooks

gregzawadzki:
ObjecModel will call hooks like

Hook::exec('actionObject'.get_class($this).'AddAfter', array('object' => $this));

Making it uncallable method name like `actionObjectthirtybees\core\testAddAfter` Normalize it to Making it callable method name like `actionObjectthirtybees_core_testAddAfter` So it can be defined as a method

### DIFF
--- a/classes/Hook.php
+++ b/classes/Hook.php
@@ -439,6 +439,9 @@ class HookCore extends ObjectModel
             $disableNonNativeModules = (bool) Configuration::get('PS_DISABLE_NON_NATIVE_MODULE');
         }
 
+        //Allow namespaced objects to execute Hooks thirtybees\core will become thirtybees_core and can be defined as method
+        $hookName = str_replace('\\', '_', $hookName);
+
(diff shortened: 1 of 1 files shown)


getdatakick:
This pull request, in its current version, do not fixes the outlined problem.

Currenlty, it only does:
```php
Tools::toCamelCase("actionObjectThirtybees\Core\TestClassAddAfter");
```
which is a no-op.

We would need to change it to something like

```php
Tools::toCamelCase(str_replace('\\', '_',  "actionObjectThirtybees\Core\TestClassAddAfter"));
```
to have a desired effect.

However, we don't want to do that either, because it would break compatibility with existing hooks that contains `_` character.

I used your implementation as an inspiration, fix is in commit 4579910f3c8003419be44dfb9290c5b41b809a44.

We don't touch existing `_`, but only replaces `\` characters in hook name with camel-case

PR #2115, opened 2026-04-24
Adding envelopeFrom support

eschiendorfer:
Adding this envelope from/return path value is helpful for setting up an email address like bounce@domain.com. Otherweise its pretty hard to cleanup your email list.

My idea was to introduce a new interface. It's a bit ugly, but on the other side it allows backward compatibility.


getdatakick:
I don't like the new interface, but I agree that it's probably the easiest and safest solution for backward compatibility

PR #2119, opened 2026-04-30
Improvement on the db backup functionality.

therampagerado:
- No race condition if we quickly open new BO page
- No exceptions in logs module
- Logging through logger

### DIFF
--- a/classes/ShopMaintenance.php
+++ b/classes/ShopMaintenance.php
@@ -58,7 +58,7 @@ class ShopMaintenanceCore
      * This method gets triggered by the 'getNotifications' Ajax request, so
      * every two minutes while somebody has back office open.
      *
-     * @throws PrestaShopException
+     * @throws Throwable
      */
     public static function run()
     {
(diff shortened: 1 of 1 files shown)


classes/ShopMaintenance.php
```diff
@@ -183,26 +207,44 @@ public static function cleanOldThemeCacheFiles()
...
+
+                return false;
             }
+
+            Configuration::updateGlobalValue(static::LAST_BACKUP_CONFIGURATION, time());
+
+            return true;
+        } catch (Throwable $e) {
+            static::logBackupFailure('Automatic backup failed: ' . $e->getMessage());
+
```
getdatakick, line 245: I'm not a big fan of swallowing exceptions. It makes it much harder to detect and find problems.

I suggest we re-throw the exception here, so it is handled by error log system.
It the exception somehow blocks the page, we should at least forward the exception to ErrorHandler.

getdatakick (changes requested):

Issue #2121, opened 2026-05-08
Customer group saving is broken for new customers


eschiendorfer:
This commit https://github.com/thirtybees/thirtybees/commit/95a1c6c94e0d9dfa24715dbe7f2f0d6f47c837c9 has broken the standard group selection when you add a new customer in AdminOrders&addorder. As no customer exists, obviously this does trigger:

    if (!idCustomer) { return; }

Also I am not sure, if we really go the window.i18n_ way. Thats at least nothing that was used in the past to translate strings.

Thanks @therampagerado. He has already fixed it here:
https://github.com/thirtybees/thirtybees/pull/2118/changes/07c9f0b2dbfc27ffb309f62794cd36b609bfe384

My personal impression is, that the ajax endpoint is a bit overkill, that's why I bring up a simpler solution:

PR #2122, opened 2026-05-08
Fixing issue https://github.com/thirtybees/thirtybees/issues/2121

eschiendorfer:
Fixing issue https://github.com/thirtybees/thirtybees/issues/2121 and removing introduced ajax endpoint.

IMO the new ajax endpoint is a bit overkill. But maybe it's a matter of taste!?

@therampagerado not sure, if you have time and if it's easy for you to test my PR. Does this work for you and does this still solve your initial main goal or is the ajax save of any importance?

Btw which model do you use, to write your code? My PR was written by codex 5.5.

### DIFF
--- a/controllers/admin/AdminCustomersController.php
+++ b/controllers/admin/AdminCustomersController.php
@@ -748,6 +748,7 @@ class AdminCustomersControllerCore extends AdminController
                 break;
             }
         }
+        unset($input);
 
         foreach ($groups as $group) {
             $this->fields_value['groupBox_'.$group['id_group']] =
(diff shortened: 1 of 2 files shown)


therampagerado:
Thank you @eschiendorfer ,

Will test it later this weekend.

I use codex 5.5 locally on high - and this was the solution it managed to spit out :). The original commit was probably 4.x

Also one addition  - when we create customers via AdminCustomers&addcustomer we should not see and be able to set Visitor in the Default customer group.

image

Similar issue I noticed in AdminOrders&addorder - we currently filter and show only customer. But what if we want to create a guest account for this order?

image

PR #2128, opened 2026-06-12
Harden store-credit code redemption for gift-card use

Smiles-v:
### Commit 1 — Remove duplicated `getDisplayPaymentOptions()`

bb4549d declared the method twice in `ParentOrderControllerCore` (identical bodies) — a PHP fatal on every page using this controller. Kept the first declaration.

### Commit 2 — Harden store-credit code redemption

The current claim flow in `ParentOrderController::init()` (`submitAddDiscount`) does:

```php
$storeCredit->id_customer = $customerId;
$storeCredit->update();
```

unconditionally — anyone who knows (or guesses) a code can re-bind credit that already belongs to another customer's account, at any time, repeatedly. The validity window and remaining balance aren't checked either, producing confusing "you don't have store credit" failures later in the flow.

Changes:

- `StoreCredit::claimForCustomer(int $idCustomer): bool` — binding is only allowed for **unowned** credits (`id_customer = 0`, i.e. gift cards waiting to be redeemed by whoever received the code) or credits already owned by the claimant.
- `StoreCredit::isCurrentlyValid()` and `getRemainingAmount()` helpers.
- The checkout claim block now returns distinct errors (*already redeemed / no longer valid / no remaining balance*) before enabling `use_store_credit`; plus a typo fix in the sign-in message.

### Why — the gift-card use case

With this in place a gift-card module can issue an **unowned** credit carrying a code: the recipient claims it at checkout and core spends it **as payment** (`BOTH_WITHOUT_STORE_CREDIT`), so the VAT base of the redeeming order stays intact — exactly the EU multi-purpose-voucher treatment — with native partial use via `amount_used` and no `-2` leftover chains. The `tbvoucherproduct` module ships a feature-detected StoreCredit issue mode on its side (v0.6.0, branch `smiles/exploration`).

One observation while in the area: `store_credit` has no currency column (implicitly the shop default) — fine for single-currency shops, worth a thought for multi-currency.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/classes/StoreCredit.php
+++ b/classes/StoreCredit.php
@@ -119,6 +119,62 @@ class StoreCreditCore extends ObjectModel
     }
 
     /**
+     * Remaining (unspent) amount on this credit.
+     *
+     * @return float
+     */
+    public function getRemainingAmount(): float
+    {
+        return max(0.0, (float)$this->amount - (float)$this->amount_used);
+    }
+
+    /**
+     * Whether the credit is within its validity window. Empty bounds
+     * (normalized to null in the constructor) mean "no restriction".
+     *
+     * @return bool
+     */
+    public function isCurrentlyValid(): bool
+    {
+        $now = date('Y-m-d H:i:s');
+        if ($this->date_from && $this->date_from > $now) {
+            return false;
+        }
+        if ($this->date_to && $this->date_to < $now) {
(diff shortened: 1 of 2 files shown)


getdatakick:
merged

PR #2130, opened 2026-07-03
Persist store-credit spending per order

Smiles-v:
Follow-up to #2128, verified against wip HEAD `6840511`.

## Problem

Paying with store credit only lowers the payable total in `Cart::getOrderTotal()`. Nothing is persisted at order creation:

- `amount_used` never increases, so the balance never decreases and the same credit lowers the next order again;
- no credit-to-order association exists;
- no `OrderPayment` row is written, so the order and its documents show nothing and the totals do not add up.

## Change

- **New `StoreCreditSpend` model** (`store_credit_spend`): one row per credit consumed by an order, with a `date_reverted` zero-date sentinel. Lookups by order and by credit.
- **`StoreCredit::spendForOrder()`**: debits the customer's valid credits, soonest expiry first. Each debit is a conditional `UPDATE ... WHERE (amount - amount_used) >= :take`, so concurrent checkouts cannot spend the same money; a lost race re-reads the balance and retries once. All or nothing: on shortfall every debit of the call is reverted and order validation aborts. Compensation failures are logged.
- **`PaymentModule::validateOrder()`**: orders paid (partly) with credit carry GROSS totals; the credit is surfaced as a regular `OrderPayment` row ("Store credit", transaction id = the consumed codes) next to the gateway payment, so payments and totals reconcile everywhere with zero template changes. The spend runs after all orders of the cart are created, inside a guarded block that restores the credit before rethrowing on any later failure.
- **`OrderHistory::changeIdOrderState()`**: credit spent on an order that lands in an error or cancelled state is restored automatically (idempotent, claim-before-refund), mirroring the stock re-injection. Refund states stay merchant-controlled on purpose.
- **`Cart::getOrderTotal()` credit branch**: floors the applied amount at the currency display precision (sub-cent balance dust no longer aborts the checkout) and only applies credit to default-currency carts, because credits carry no currency.
- `StoreCredit::getByCustomerId()` reads from master (the figure decides how much money a checkout may spend).
- BO: read-only spend history (date, order link, amount, spent/reverted) on the credit edit form.
- Second commit fixes pre-existing swapped `(shopId, customerId)` arguments in both `ParentOrderController` balance lookups.

## Testing

Functionally tested against a live database: consumption order, read-back, shortfall revert, restore on cancellation incl. idempotence, external compensation, and the conditional-debit race guard (second identical debit affects zero rows). Schema is definition-driven; both models added to `WebserviceMetadataTest::NOT_EXPOSED`.

## Known limitations (documented, out of scope)

- Concurrent duplicate `validateOrder()` for one cart (pre-existing TOCTOU on `orderExists()`) can create two orders and legitimately debit twice.
- All-virtual carts bypass the credit branch (`BOTH` is rewritten to `BOTH_WITHOUT_SHIPPING` before it).
- Automatic restore on refunds/credit slips is a suggested follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/classes/Cart.php
+++ b/classes/Cart.php
@@ -1115,10 +1115,20 @@ class CartCore extends ObjectModel
             return $orderTotalDiscount;
         }
 
-        if ($orderTotal > 0 && $this->use_store_credit && (int)$this->id_customer) {
+        // Store credits carry no currency, so they are only applied to carts
+        // in the shop default currency; a JPY cart would otherwise drain a
+        // EUR-granted balance one-to-one.
+        if ($orderTotal > 0 && $this->use_store_credit && (int)$this->id_customer
+            && (int)$this->id_currency === (int)Configuration::get('PS_CURRENCY_DEFAULT')) {
             if ($type == static::BOTH || $type == static::ONLY_STORE_CREDIT) {
                 $creditAvailable = StoreCredit::getByCustomerId((int)$this->id_shop, (int)$this->id_customer);
-                $creditUsed = Tools::roundPrice(max(0.0, min($orderTotal, $creditAvailable)));
+                // FLOOR to the display precision. The credit balance can hold
+                // sub-precision dust (6-decimal column); rounding UP would
+                // promise more than the balance covers and make the order-time
+                // debit fail deterministically. Flooring leaves the dust on
+                // the credit instead of blocking the checkout.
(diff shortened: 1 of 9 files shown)


getdatakick:
Merged via separate commit. I had to made few changes, like replacing inlined html with theme template.

PR #2136, opened 2026-07-16
Unit price fix and multilanguage field for it.

therampagerado:


getdatakick:
Thank you Rado for the PR. 

Unfortunately I won't be able to merge this, because we are in process of implementing units differently - standalone units definition in separate records, with conversion rates logic, etc.

therampagerado:
> Thank you Rado for the PR.
> 
> Unfortunately I won't be able to merge this, because we are in process of implementing units differently - standalone units definition in separate records, with conversion rates logic, etc.

Great! Thanks!

PR #2139, opened 2026-08-26
Fix bool list filter showing dash when filtering on No

Smiles-v:
Bool column filters are deserialized to real PHP booleans (BoolValueType::deserializeValue), so a No filter reaches list_header.tpl as false. The No option's selected check required 'value != empty string', and false loosely equals empty string in PHP, so the option was never marked selected: the list was correctly filtered on No while the dropdown displayed the unfiltered dash. Affects every COLUMN_TYPE_BOOL filter in every admin list. Fixed by comparing strictly against the boolean, int and string representations for both options; the dash still shows for null/empty (no filter). Found while testing the WMS fulfilment list Paused/Batch filters.


getdatakick (commented):
Looks good

